Issue #136, opened 2015-01-23
Improve state synchronization between server and client.


koo-taejin:
purpose
1. to use duplex communication more stable. 
2. to save agent state more detail. 

to-do (not fixed)
1. integrate socket state code. 
2. create socket state id. 
3. improve health check.

PR #221, opened 2015-03-10
Improve state synchronization between server and client. #136

koo-taejin:
1. Integrate SocketState class into PinpointSocketHandler.


jaehong-kim:
if for refactoring, ok

Issue #215, opened 2015-03-06
Sending server meta data should be separated from the agent's lifecycle


Xylus:
The agent's lifecycle (and sending agent information) is currently controlled by plugins. Each plugin must specify if it will control the agent's lifecycle, and call `Agent.start()` manually if it does.

For example, a Tomcat 6 application is modified to invoke `Agent.start()` after `org.apache.catalina.core.StandardService.start()` method completes, and only then would the agent sends its information along with the server's meta data.

The agent's information should be sent regardless of the enclosing application's status, and this requires adding a way for the application to notify that it has finished gathering the server's meta data.

Xylus:
Making further modifications, It would be better for `PinpointBootstrap` to govern the agent's lifecycle (ie. start/stop methods).

PR #222, opened 2015-03-10
separate sending of agent information and server meta data - #215

Xylus:
Added an API for the plugin interceptors to invoke once the application startup process is finished.

DefaultServerMetaDataHolder's publishServerMetaData() method iterates through the registered listeners and passes the current ServerMetaData object to them.


Xylus:
Closing to make further modification - Move agent start/stop timing to `PinpointBootstrap` (from `DefaultAgent`)

Issue #246, opened 2015-03-23
Dockerizing Pinpoint


nstopkimsk:
We would like to dokerize Pinpoint, any  Docker experts out there?

jcrygier:
Hi...I had a go at this, but I'm no expert on Pinpoint / HBase / ZooKeeper, so I fell short on some of the networking bits.  I tried to break them out into 3 images: HBase, Collector, and Web.  Feel free to take a peek at my work (ignore the directories that don't start with pinpoint):

https://github.com/Ventiv-Technology/DockerManager/tree/master/sample-config/dockerfiles

hoonmin:
@jcrygier Hi. Thank you for sharing!
I took a peek your sample and with small fixes I could run it successfully.
- invalid templates on `hbase.properties`: missing braces...
- currently, pinpoint-web should be on `webapps/ROOT`

So, could you fix it and send us a PR to manage it in the pinpoint project?
Then we can push them to docker hub as official pinpoint images.

Xylus:
:+1:

jcrygier:
Alright, I've fixed the templates for environment variables, and moved collector / web to run under ROOT.  Please review PR attached.

NOTE: I still was unsuccessful at running, but this looked much better than it did before.  I'm unsure if there is an issue with my test application...I will try to use the sample you provide when I get some free time.

nstopkimsk:
jcrygler contributed this issue.

PR #283, opened 2015-03-31
Dockerize HBase, Collector and Web

jcrygier:
Initial attempt at Dockerizing pinpoint, per Issue #246.


hoonmin:
Hi. I'm sorry to late response for your PR.
Today I tested your dockerfiles and built a fully-functioning pinpoint environment with `a tiny modification`.

Would you take a look for it?
https://github.com/jcrygier/pinpoint/pull/1

I used `--net=host` option to make HBase accessable easily. (HBase looks for hostnames, not IP :-( )
But if you don't like it, you can link the pinpoint-collector and pinpoint-web with pinpoint-hbase. Docker should add the hbase hostname to `/etc/hosts` automatically.

Thanks!

jcrygier:
I've accepted your changes...as it's simply in the comments anyway.  However, I think it's still preferable to link, rather than use the host's networking.  That seems like a bit of a security concern to me.

But again, since these are just comments, it really has no effect.  It's still up to the end user to decide what is best for their environment.

hoonmin:
Good. Thanks for your contribution!

PR #376, opened 2015-04-28
Optimize udp efficiency.

koo-taejin:
1. create span stream protocol
2. change udp socket to datagram channel
3. reuse send buffer by gathering channel


jaehong-kim:
good

PR #387, opened 2015-05-04
Add @suppressWarnings("deprecation") for using DEFAULT_CONTENT_CHARSET i...

pavian:
For "DEFAULT_CONTENT_CHARSET in org.apache.http.protocol.HTTP has been deprecated" compile warning, I added "@SuppressWarnings("deprecation")".

For "NoSyncConnected in org.apache.zookeeper.Watcher.Event.KeeperState has been deprecated" compile warning, I removed the case checking whether keeperstate is NoSyncConnected because NoSyncConnected is deprecated and said that it will never be generated by server


koo-taejin:
Old version zookeeper server  still makes NoSyncConnected Event.
So I can't accept you're pull request. 

We want to support zookeeper server regardless of version as possible. 

Thanks.

pavian:
I just add @supressWarnings to zookeepers, not removing NoSyncConnected.

Issue #444, opened 2015-05-19
Refactoring naming and folder structure according to convention


denzelsN:
https://github.com/johnpapa/angular-styleguide

PR #445, opened 2015-05-19
Refactoring naming and folder structure according to convention

denzelsN:
https://github.com/johnpapa/angular-styleguide


denzelsN:
https://github.com/naver/pinpoint/issues/444

PR #459, opened 2015-05-22
Fix syntax of markdown files

yous:
Inline code can be written with one backtick `` `code` ``. Three backticks are used for multiline code:

    ```
    code
    ```


nstopkimsk:
thanks.

PR #711, opened 2015-07-13
improve trace event ordering

Xylus:
- SpanEvents now uses its wrapping Span's id when ordering
- Asynchronous events are now guaranteed to be placed after its parent
event (used to cause problems when async traces happened at the same
time as its parent event)
- Fix cases where nulls are passed into annotations when creating
ExpectedTrace (cases where ExpectedAnnotation[] is used instead of
varargs)


Xylus:
closing, fixed in #721

Issue #84, opened 2015-01-08
Add lifecycle support for agents


Xylus:
Agent life cycle support is currently missing. 
Collectors should be able to keep track of the life cycles of connected agents.

Additionally collector/web should be able to keep track of various events generated by agents (including life cycle events, and events generated by users from the Web UI)

Xylus:
Basic implementation complete. Features branching from life cycle / event support will have their own issues created.

PR #780, opened 2015-07-28
add data handling of agent life cycles and events

Xylus:
Add agent life cycle and event handling support mentioned in #84 to the Collector.This pull request only includes the receiving and storing part of the agent life cycle / event data, retrieval and querying for these data is not included yet.

2 tables were added to the HBase backend - one for agent life cycle, and one for agent events.

This implementation tracks the agent's connect / ping / disconnect events to determine the latest status of a single life cycle of the agents.
These life cycle events are also put into the event table to provide a separate history of events generated to/from the agent.


koo-taejin:
Looks great! :)

Xylus:
Thanks!

PR #820, opened 2015-08-10
Refactor route handling to allow wiring of route filter beans

Xylus:
Refactored to allow route filters to be registered as Spring Beans and have them wired into route handlers.
This will allow the event handler to be wired into a route filter to handle command events in relation to #84


koo-taejin:
pretty good :)

Xylus:
thanks!!!

PR #966, opened 2015-09-16
The Servlet parameter provides an option for the profile.

youtome:


lioolli:
Tomcat request is not profiled by HttpServletInterceptor. It will be removed from 1.5.

emeroad:
Refer to the following code.
https://github.com/naver/pinpoint/blob/1.1.x/profiler/src/main/java/com/navercorp/pinpoint/profiler/modifier/tomcat/interceptor/StandardHostValveInvokeInterceptor.java
line:202

youtome:
Ok. I got it and I will be modify StandardHostValveInvokeInterceptor.java.
Please close this request.

PR #1092, opened 2015-10-20
Add depth, period, favorite configuration.

denzelsN:


SmileAngel007:
is it available to modify the follow files? thank you very much ,i want to modify the depth too~~~

PR #1126, opened 2015-10-28
just fixing typos in Comments and Log messages

yjaeseok:
just fixing typos in Comments and Log messages


koo-taejin:
Thanks!!!!

Issue #1254, opened 2015-11-24
File /images/icons/JETTY.png does not exists


dawidmalina:
File /images/icons/JETTY.png does not exists.

Xylus:
Good catch! We've also got `/images/servermap/JETTY.png` missing :)

Xylus:
Merged and done, thanks again!

PR #1255, opened 2015-11-24
File /images/icons/JETTY.png does not exists #1254 

dawidmalina:
https://github.com/naver/pinpoint/issues/1254


Xylus:
@dawidmalina Thank you for this. :+1: I've got 2 requests for you!
First is the 827fb30 commit up there - This might be a non-issue once merged but I'm not entirely sure how it'll look. Could you rebase and throw us a PR again?
Second is we also got `/images/servermap/JETTY.png` missing. This one's used for the server map. If you could whip up a similar one here, that'd be great!

dawidmalina:
Sure I will create new (clean) PR.

Xylus:
Looks great, thanks again!
:+1:

PR #1378, opened 2015-12-24
Add ability to build docker images through mvn (via docker-maven-plugin)

dawidmalina:
With this we can very easily build local changes as docker images
```
mvn docker:build
```

and run pinpoint dockerized via docker-compose
```
docker-compose up -d
```

To make this happen I've added:
- new maven plugin
- create fake hbase module
- move scripts folder to hbase module

This approach may look weird at the beginning but it is very helpful during testing. Dockerizing is not bound into regular maven phase so will not expand build time unless someone run this `mvn docker:build`


dawidmalina:
To work properly (agent attached as docker volume) we need this change to be merged and released https://github.com/spotify/docker-maven-plugin/pull/161

nstopkimsk:
@dawidmalina thanks. We need more time to review and test this.

nstopkimsk:
LGTM!!  but other guy is checking some other side-effects. 
It's very simple way to test and quickstart pinpoint.

PR #1384, opened 2015-12-30
Refactor SpanAligner - remove repeated span logs for async traces

Xylus:
Repeated logging of Spans could generate huge log files for transactions with thousands of asynchronous traces. This change only logs Spans once for each span.


jaehong-kim:
good.

Issue #1430, opened 2016-01-12
Improve Realtime Active Thread Chart smoother.


koo-taejin:
Improve Realtime Active Thread Chart smoother.

PR #1431, opened 2016-01-12
Improve Realtime Active Thread Chart smoother. #1430

koo-taejin:
Improve Realtime Active Thread Chart smoother.


koo-taejin:
@emeroad Could you please review this PR.   :)

PR #1454, opened 2016-01-15
fix invalid classLoading

emeroad:
- support spring aspectj weaving

ping @lioolli  @jaehong-kim


jaehong-kim:
reviewed-by @jaehong-kim

PR #1463, opened 2016-01-19
change logic from recursive to loop, because of avoid call-stack-over…

jaehong-kim:
#1448

@emeroad


emeroad:
reviewed-by @kang

Issue #1478, opened 2016-01-22
Improve logger logic for pinpoint core module


minwoo-jung:
correlate to each other #1413

minwoo-jung:
reviewed by @emeroad

PR #1486, opened 2016-01-26
[#1478] add JdkLoggerFactory to prevent using slfj4 logger by jul-to-slf4j library

minwoo-jung:


minwoo-jung:
please review @emeroad

Issue #1482, opened 2016-01-25
[PinpiontWeb] fix a ActiveThread concurrency problem


koo-taejin:
If multiple threads invokes WebSocketSession's sendMessage method at the same time, ; 
then it is possible to makes IllegalStateException.

PR #1499, opened 2016-01-29
[PinpiontWeb] fix a ActiveThread concurrency problem #1482

koo-taejin:
1. added OrderedThreadPoolExecutor.
2. separated thread from WebSocketSessionId.


koo-taejin:
ping @emeroad @Xylus

Issue #1443, opened 2016-01-14
Fixed invalid classloading when using the AspectJ + Spring AOP


emeroad:
- support spring aspectj weaving
- support OSGI  (atlassian jira)
 -   #1583  Enhancement PlaingClassLoaderHandler for Jira.

PR #1500, opened 2016-01-29
Fixed invalid classloading when using the AspectJ + Spring AOP #1443 

emeroad:
#1443 
- refactoring api
- add testcase


emeroad:
ping @jaehong-kim

PR #1510, opened 2016-02-02
[#1508] Add config check logic in log4j, logback plugin

minwoo-jung:


minwoo-jung:
please review @lioolli

lioolli:
reviewd by lioolli

PR #1531, opened 2016-02-05
Show active trace chart along with agent stat data

### LINKED ISSUE #1504: Collect active thread data every stat collection interval
Realtime active thread data is useful when monitoring applications. It would be beneficial to store these data for troubleshooting past events as well.
Collection would be done by taking a snapshot of the active trace data every certain intervals, most likely along with agent stat collection shipping the data to the collector.

Xylus:
Collect active trace along with other agent statistics.
The actual chart needs to be implemented in front-end.
Resolves #1504 

ping @koo-taejin @emeroad


koo-taejin:
reviewd by @koo-taejin :+1:

PR #1535, opened 2016-02-11
[#1530] handle exception when request entity don't support getContent

minwoo-jung:
handle exception when request entity don't support getContent


minwoo-jung:
please review @emeroad

Issue #1362, opened 2015-12-18
Implement parallel result scanner for scanning rows distributed with RowKeyDistributor


Xylus:
Allow distributed scanners to run in parallel that would theoretically allow for better performance when scanning rows that are well distributed using `RowKeyDistributor`

Xylus:
Reopening as scanner creations may be done in parallel as well, and this may cut some time during scanner creation.

nstopkimsk:
NAVER - http://www.naver.com/
--------------------------------------------
김성관(sungkwan.kim@naverco...) 님은 현재 부재중입니다.</br>
보내신 메일 <Re: [pinpoint] Implement parallel result scanner for scanning rows distributed with RowKeyDistributor (#1362)> 은 저장되어 있으므로 다시 보내실 필요는 없습니다.</br>
김성관(sungkwan.kim@naverco...) 님이 남기신 메시지 입니다.
--------------------------------------------
[Absence period : 2015-12-24 10:00 ~ 2015-12-31 23:59 GMT +09:00]<br>리프레쉬 휴가입니다. <br>Pinpoint는 Pinpoint DL로 문의주시기 바랍니다. 
--------------------------------------------

Xylus:
Reopening to apply parallel scan for finding accept applications for nodes when drawing the server map.

PR #1543, opened 2016-02-18
#1362 Make scanner creation run parallel

koo-taejin:
include tableFactory object when to execute releaseTable method.


koo-taejin:
ping @Xylus :)

PR #1554, opened 2016-02-25
allow whitespace for collector.l4.ip configuration in pinpoint-collec…

koo-taejin:
…tor.properties #1553


koo-taejin:
ping @Xylus  :)

Issue #1518, opened 2016-02-03
Add method, that get scatter data made of dot group  .


koo-taejin:
strong point
1. reduce size of scatter data. 
2. draw faster for scatter chart

PR #1568, opened 2016-03-03
Add method, that get scatter data made of dot group . #1518

koo-taejin:
added serverList value in getFilteredServerMapDataMadeOfDotGroup api result.


koo-taejin:
@jaehong-kim review please :)

Issue #1573, opened 2016-03-04
Add direction option to rest apis that getting ScatterData.


koo-taejin:
Add direction option to rest apis that getting ScatterData.

PR #1574, opened 2016-03-04
Add direction option to rest apis that getting ScatterData. #1573

koo-taejin:
added backwardDirection option.


koo-taejin:
ping @Xylus :)

PR #1576, opened 2016-03-05
Upgrade Apache Commons Collections to v3.2.2

jart:
Version 3.2.1 has a CVSS 10.0 vulnerability. That's the worst kind of
vulnerability that exists. By merely existing on the classpath, this
library causes the Java serialization parser for the entire JVM process
to go from being a state machine to a turing machine. A turing machine
with an exec() function!

https://commons.apache.org/proper/commons-collections/security-reports.html
http://foxglovesecurity.com/2015/11/06/what-do-weblogic-websphere-jboss-jenkins-opennms-and-your-application-have-in-common-this-vulnerability/

Also, do consider using Guava in the future.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -518,7 +518,7 @@
             <dependency>
                 <groupId>commons-collections</groupId>
                 <artifactId>commons-collections</artifactId>
-                <version>3.2.1</version>
+                <version>3.2.2</version>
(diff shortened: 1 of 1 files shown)


emeroad:
thanks for your contribution.
We will analyze the security issues.

PR #1588, opened 2016-03-10
websocket bean id correct

vwyuheng:
websocket spring bean id correct


koo-taejin:
I think I want to reject this one ;)

PR #1590, opened 2016-03-10
add getResponseTimeHistogramData restful api. #1589

koo-taejin:


koo-taejin:
@jaehong-kim review please :)

PR #1626, opened 2016-03-19
new plugin: jdk async task trace

hamlet-lee:
This plugin traces JDK provided FutureTask.

BTW:
Currently, the FutureTask's constructor's parameter's class name is not shown in the Web UI.
What should I do to display that information in Web UI? This information helps to figure out which task it is.


hamlet-lee:
@nstopkimsk JDK_EXEC is an common SERVICE_TYPE, would you allocate a new SERVICE_TYPE for this plugin?

vishalmodak:
What is the status on this PR?

hamlet-lee:
No update to the plugin recently. Do you have any suggestion for this plugin to be merged into master?

Anishkp:
Can someone resolve the conflicts?

Anishkp:
Can this be resolved?

dinesh4747:
Any plans for merging this PR in the upcoming version 1.7.0 ?

emeroad:
We don't recommend storing strong reference
Pinpoint does not use a solution to store strong reference.
This solution has OOM risk and can cause bug.
 - eg : GlobalMap.put(traceTraget) or Timer.register(traceTraget) and GlobalMap.remove(traceTraget) 

We are not ready for Java asynchronous API (eg : ExecutorService, Future~~).

dinesh4747:
HI @emeroad 

Thanks for conforming back in time
If OOM would incur due to tracing implementation of Future and FutureTask interface then we can still get the tracing only for specific Callable interface on-demand by reducing the scope

dinesh4747:
Myself and @Xylus had a brief discussion some time back and decided to try to control the implementations over pinpoint.config as mentioned below

HyunGil Jeong @Xylus 12:04
In the asynchronous libraries we support (such as Vert.x), we modify the classes that are from that specific library. So any side effects of bytecode changes are confined within it. We know how these classes are used and methods are invoked so we can assume a lot of things when we change their bytecode.
Classes from JDK, we cannot make this assumption. We have no idea how people are going to use for example, a ExecutorService, or a Callable. If we decide to transform ExecutorService and Callable to allow trace propagation, it affects not only the application that uses them, but also any libraries that uses them as well.
So we're trying to stay away from modifying classes supplied by JDK itself for now.
dinesh4747 @dinesh4747 12:11
But still I feel that callable interfaces can be traced for propogration* and will not cause any harm to appl uses them
HyunGil Jeong @Xylus 12:15
That may very well be true! And yes, the concern might be blown out of proportions. Though, I think reducing the scope might still be necessary. While I think modifying everything that implements Callable is a bit too much (not to mention increase in class load time), I think if we could specify certain classes that implement Callable in pinpoint.config to narrow down the scope, that would be better. What do you think?
dinesh4747 @dinesh4747 12:17
yes absolutely makes sense

dinesh4747:
@Xylus - could you please share your thoughts on this?

Xylus:
@dinesh4747 
We'll probably need more ground work done before we can explore the idea further (such as matching classes to transform that implements certain interfaces?)

hamlet-lee:
@emeroad 
In CacheMap, there are two protections for OOM

1. only keep object generated in 10 secs if more than 100 items associated with the thread
1. when adding object to a thread's cacheMap, we check at most 100 items can be kept

With upper protections, I think OOM risk is very low.

And we can also make 100 as configurable in the future, which will also lower the risk of OOM.
Anyone don't want to have any risk on this plugin, can also shut it down in configuration too.

dinesh4747:
Hi @hamlet-lee 

I would strongly share your view here and we can avoid the risk of OOM occurrence by loading with the following:

1) Limited load factor 

2) Minimize the TTL period to ensure that unused objects are being evicted on periodical basis with bounded expiry time 

@Xylus @emeroad - Kindly share your thoughts on this

dinesh4747:
Hi All - @hamlet-lee @emeroad @Xylus

I would like to hear your valuable thoughts on my previous comments, also would like to know like whether any potential traction to wrap this PR soon ?

hamlet-lee:
Current implementation have oom protection in design:
There are two kind of data structure in memory which could expand.
1. a global map:   CacheMap.getInstance(JdkExecConstants.ASYNC_ID_MAP).sharedMap
2. a map for each worker thread: CacheMap.getInstance(JdkExecConstants.ASYNC_ID_MAP).threadLocalCache.get()

For global map:
1) we only keep simple object in map.
```
static class ValuePair{
        Object val;  //not really an object, we just store object's hash code (int)
        long ts;
        long threadId;
}
```

2) we check size of map and age of map in a periodic way.
current implement is every 10 secs we check its size and age of items.
```
+            new Timer("pinpoint-JDK-EXEC-CacheMap-"+name+"-cleaner").schedule(new TimerTask(){
 +                @Override
 +                public void run() {
 +                    //auto clear every 10 secs
 +                    if( sharedMap.size() < 100) {
 +                        //do nothing if the map is not big
 +                        return;
 +                    }
 +                    long maxAge = System.currentTimeMillis() - 1000*10;
 +                    for (Map.Entry<Object, ValuePair> entry : sharedMap.entrySet()){
 +                        if(entry.getValue().ts < maxAge)
 +                        {
 +                            sharedMap.remove(entry.getKey());
 +                        }
 +                    }
 +                }
 +
```
For thread local map:
1) we only keep simple object in map, same as the items in global map
2) we only store item when running the related async task.  The item is stored for future read.
3) we limit the size of the map when adding item to it
```
+            //size check
 +            if( arrayList.size() > 100) {
 +                //nested 100 layers, should never happen
 +                if( logger.isDebugEnabled() ) {
 +                    logger.debug("local cache size too large," +
 +                            " something happened!", new Throwable());
 +                }
 +                arrayList.clear();
 +                hashMap.clear();
 +            }
```
4) we aggressively remove items from the map: if older task's meta data is being getting 2nd time(when task ends), but newer task's meta data is not read (when newer task ends).
This won't happend in normal logic.  The logic is just for safety.

So, wrap up.
1) no object with out reference is stored (just longs and ints) memory footprint is relatively small.
2) size of local thread map is strictly limited (no more than 100 items)
3) size of global map's size and ttl is checked each 10 second.  The reason why we do not use a hard limit on its size, is because:
The customer's service may generate many async tasks in an instant (e.g. 10000/sec) and pending it for some seconds and then run and finish the asyncs tasks in seconds.
The global map should hold metadata for these 10000 pending async tasks.  The number 10000 may varied between systems which are running well, so I do not see obvious hard limitation value can be specified.
Good news is that, the global map only keep manageable & reasonable data in memory, its size is just O(n), n is pending tasks generated in 10 secs (checking period).

I do not have much time to catch up with pinpoint's evolution or follow the rules of contribution.
So, I am glad if anyone can take over this PR and make it merged into main stream.

hamlet-lee:
no update

tankilo:
Will there be any new progress? It seems the author signed the CLA this July.  
This feature deserves attention :)    
@hamlet-lee @Xylus

tankilo:
For some reason, i've modified this pull request to pinpoint 1.7.3. 
The way to trace async thread has changed.
Refer to [WorkerConstructorInterceptor.java](https://github.com/naver/pinpoint-plugin-sample/blob/master/plugin/src/main/java/com/navercorp/pinpoint/plugin/sample/_12_Asynchronous_Trace/WorkerConstructorInterceptor.java), i need to add field with type `AsyncContextAccessor` to `java.util.concurrent.FutureTask`.

image

image

I find that this pull request use `java.lang.instrument.Instrumentation#retransformClasses` to retransform `java.util.concurrent.FutureTask`,  the javadoc of this method states the following restriction.
> The retransformation must not add, remove or rename fields or methods, change the signatures of methods, or change inheritance.

`com.navercorp.pinpoint.profiler.instrument.ASMClass#addField` violate it, so the exception is raised:( 
image

1103036128:
Any updates on the PR ? @hamlet-lee @Xylus @emeroad
I am using version 1.8.4

PR #1636, opened 2016-03-21
Modified Hbase version to 1.0.3 to fix docker:build error.

nstopkimsk:
There  is no Hbase 1.0.2 from apache.mirrors.pair.com/hbase.


nstopkimsk:
ping @koo-taejin

PR #1651, opened 2016-03-23
modified configurations for docker to show realtime active thread chart correctly

nstopkimsk:
1. modified configurations for docker to show realtime active thread chart correctly using docker compose.

2. modified hbase version of quickstart app.


nstopkimsk:
ping @koo-taejin

PR #1694, opened 2016-04-15
Support the async put API provided by HTableMultiplexer to improve th…

koo-taejin:
…e write throughput in collector. #1683

 - If asyncOperation is not set, then execute put method instead of asyncPut method.
 - wait destory method until async queue is empty.


koo-taejin:
@emeroad review please :)

emeroad:
reviewed-by emeroad

Issue #1701, opened 2016-04-19
Consider separating DatagramSocket from UdpDataSender class.


koo-taejin:
Consider separating DatagramSocket from UdpDataSender class.

It allows UdpDataSender to use a specific Socket(Ex : DatagramSocket, DatagramChannel).
after, improving efficiency of data transfer to use function of specific Socket. (Ex : gathering write, DirectBuffer, ..)

koo-taejin:
discard this issue.

PR #1702, opened 2016-04-19
Consider separating DatagramSocket from UdpDataSender class. #1701

koo-taejin:


koo-taejin:
@emeroad  review please :)

Issue #1732, opened 2016-04-29
the number of success and failure dot in scatter chart different from success and failure dot count .


koo-taejin:
![2016-04-29 14 31 07](https://cloud.githubusercontent.com/assets/10057874/14908815/21af8a60-0e17-11e6-8f2b-d543d30354b4.PNG)

PR #1733, opened 2016-04-29
the number of success and failure dot in scatter chart different from…

koo-taejin:
… success and failure dot count . #1732


koo-taejin:
@jaehong-kim review please :)

PR #1736, opened 2016-05-01
Update QuickStart README to use HBase 1.0.3

yous:
This is a follow-up of #1651.


nstopkimsk:
@yous thanks for your contribution

PR #1744, opened 2016-05-04
Support the async put API provided by HTableMultiplexer to improve th…

koo-taejin:
…e write throughput in collector. #1683

 - support to reuse connection
 - added HBaseAsyncOperationMetric to CollectorMetric
 - show  HBaseAsyncOperationMetric to use JMX


koo-taejin:
@emeroad review please :)

PR #1747, opened 2016-05-09
allow mapper/*Mapper.xml in all jars classpath for MyBatis. #1746

koo-taejin:


koo-taejin:
@emeroad review please :)

PR #1775, opened 2016-05-17
extract tools module #1774

koo-taejin:
extract tools mudule.


koo-taejin:
@emeroad  please review :)

Issue #1804, opened 2016-05-26
added executor made of disruptor.


koo-taejin:
added executor made of disruptor (https://github.com/LMAX-Exchange/disruptor).

Since it enables user to choose with option where executor in place of it.

koo-taejin:
There is no performance improvement in test results.
So delete all of the code that is related to disruptor.

PR #1810, opened 2016-05-26
added executor made of disruptor. #1804

koo-taejin:


koo-taejin:
@emeroad review please :)

Issue #1822, opened 2016-05-31
Add maven profile to separate local mode (development) and release mode. 


koo-taejin:
All configuration will be changed that is much lighter than current configuration.

* local mode (development mode)
 The local mode is relatively makes many log. 

* release mode
 The release mode is relatively makes small log.

PR #1823, opened 2016-06-01
Add maven profile to separate local mode (development) and release mo…

koo-taejin:
…de. #1822

All configuration will be changed that is much lighter than current configuration.

local mode (development mode)
The local mode is relatively makes many log.

release mode
The release mode is relatively makes small log


koo-taejin:
@emeroad review please :)

PR #1833, opened 2016-06-09
Animal Sniffer Plugin applied for Java compatibility test

999fg:
- Now the plugin tests whether APIs used for pinpoint are compatible for java 1.6.
- Two files are modified: pom.xml, web/pom.xml
- This test automatically runs during project build. (ex command/ mvn install -Dmaven.test.skip=true)

***pinpoint-web is tested if it is compatible with java 1.7
***sun libraries are ignored at the test.

https://github.com/naver/kaist-oss-course/issues/65

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -77,7 +77,7 @@
         <spring-batch-version>2.2.7.RELEASE</spring-batch-version>
         <docker.maven.plugin.version>0.4.3</docker.maven.plugin.version>
         <cassandra.driver.version>2.1.7.1</cassandra.driver.version>
-
+        <sniffer.artifactid>java16</sniffer.artifactid>
     </properties>
 
     <dependencies>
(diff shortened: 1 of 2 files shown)


Xylus:
@999fg 
Looks good!
One possible suggestion would be to pull [sniffer artifactId](https://github.com/naver/pinpoint/pull/1833/commits/6940b52bac9b412e52f8c7f301ba318b366ab2bf#diff-600376dffeb79835ede4a0b285078036R895) out as a maven property.
This way, web's *pom.xml* could simply overwrite this property to `java17` instead of redefining the whole plugin. (Similar to how the *maven-compiler-plugin* is using `jdk.version` to configure source and target levels)

Automatically setting sniffer's artifactId based on `jdk.version` would be even better, but I'm not sure if this can be done easily.

I'd love like to hear what you think, please let me know.
Thanks!

999fg:
@Xylus 

Thanks for the comment!
I totally agree with your suggestion, which is more good way of coding and compatible in further development.
When I sent this PR few hours ago, I personally did not like two lines of log appearing when pom-web is being compiled (possibly due to re-defined plugin application in two pom.xml file). Following your suggestions solved this problem.
Anyway, I updated the commit as your suggestion.
Thanks!

PR #1891, opened 2016-07-04
Regression introduced with #1804

dawidmalina:
Regression introduced with commit "added executor made of disruptor. #1804"


emeroad:
@koo-taejin

koo-taejin:
@dawidmalina thanks  !!

Issue #1883, opened 2016-06-30
added to insert a number of agent into rdb periodically.


koo-taejin:
added to insert a number of agent into rdb periodically.

PR #1897, opened 2016-07-08
added to insert a number of agent into rdb periodically. #1883

koo-taejin:


koo-taejin:
already reviewd by https://github.com/naver/pinpoint/pull/1894

PR #1960, opened 2016-08-04
[#1959] add websocket data filter interface

minwoo-jung:


minwoo-jung:
please review @koo-taejin

Issue #1990, opened 2016-08-18
improve collector's stateReport.


koo-taejin:
1. Added monitored item.  
 * as-is
 ```
Pinpoint-UDP-Span-rejected
Pinpoint-UDP-Span-timer
Pinpoint-UDP-Stat-rejected
Pinpoint-UDP-Stat-timer
```

 * to-be
 ```
Pinpoint-UDP-Span-Worker.submitted
Pinpoint-UDP-Span-Worker.rejected
Pinpoint-UDP-Span-Worker.completed
Pinpoint-UDP-Span-Worker.running
Pinpoint-UDP-Span-Worker.dispatchDuration
Pinpoint-UDP-Span-Worker.duration

Pinpoint-UDP-Stat-Worker.submitted
Pinpoint-UDP-Stat-Worker.rejected
Pinpoint-UDP-Stat-Worker.running
Pinpoint-UDP-Stat-Worker.completed
Pinpoint-UDP-Stat-Worker.dispatchDuration
Pinpoint-UDP-Stat-Worker.duration

TCPReceiver-Worker.submitted
TCPReceiver-Worker.rejected
TCPReceiver-Worker.running
TCPReceiver-Worker.completed
TCPReceiver-Worker.dispatchDuration
TCPReceiver-Worker.duration
```

2. Enables to set individual
 ```
collector.tcpWorker.monitor=true/false
collector.udpStatWorker.monitor=true/false
collector.udpSpanWorker.monitor=true/false
```

PR #1993, opened 2016-08-18
improve collector's stateReport. #1990

koo-taejin:


koo-taejin:
review please @jaehong-kim

Issue #2016, opened 2016-08-29
Add ASM bytecode framework.


jaehong-kim:
Add ASM(http://asm.ow2.org/)

pinpoint.conf
~~~
# Allow bytecode framework
profiler.instrument.asm=true/false
~~~

PR #2017, opened 2016-08-29
add asm instrument

jaehong-kim:
#2016

### DIFF
--- a/agent/src/main/resources-local/pinpoint.config
+++ b/agent/src/main/resources-local/pinpoint.config
@@ -76,6 +76,9 @@ profiler.callstack.max.depth=64
 # weather or not to propagate exceptions occurred at interceptor
 profiler.interceptor.exception.propagate=false
 
+# Allow bytecode framework
+profiler.instrument.asm=false
+
 # bytecode dump option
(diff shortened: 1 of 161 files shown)


jiaqifeng:
@jaehong-kim 
I tried to read the code while it was really many! I just want to know can I instrument all sub-class of a specified base class?
I contributed the hystrix plugin. In hystrix, user inherit from HystrixCommand and implement the abstract method run(). This run() method should be intercepted but pinpoint previous release could only inspect a class you know its class name. All I want is to intercept run() method of all sub-class of HystrixCommand.
Can this PR support what I want? If not, will you consider support it in the future?

jaehong-kim:
Not yet supported.
Performance is still difficult to check for inheritance between classes.

But someday we will support it.

jiaqifeng:
@jaehong-kim 
So you use ASM to replace Javassist just to improve performance? I have seen another APM using ASM and support what I want. While search in inheritance really will slow the boot time a lot. But you know, most apps boot time exceed a half minutes, it is not acceptable to double the boot time. ^-^

Issue #2025, opened 2016-08-30
 Allows send data that consist of Span and SpanChunk List in a specific situation.


koo-taejin:
If Span and SpanChunk have less SpanEvent,
then Allows send data that consist of Span and SpanChunk List.

PR #2028, opened 2016-08-30
#2025 Allows send data that consist of Span and SpanChunk List in a s…

koo-taejin:
…pecific situation.


koo-taejin:
please review @emeroad

koo-taejin:
I'll send it again after 1.6.0 release.

PR #2084, opened 2016-09-09
[#2083] add pugin enable option and control exception for parsing jdbc url

minwoo-jung:


minwoo-jung:
please review @emeroad

PR #2109, opened 2016-09-27
Adds tracing for HystrixCommand fallback method invocation and support for hystrix-core 1.5.4+

Xylus:
This mainly adds 2 features:
1. Adds tracing for when `HystrixCommand.run()` fails and the fallback method is invoked
2. Fixes issues with transforming `HystrixCommand`'s anonymous inner class in hystrix-core 1.5.4+
### Tracing fallback methods

2 new annotations have been added to separate between `HystrixCommand.run()` and `HystrixCommand.getFallback()` invocations.
- `hystrix.command.execution` (run | fallback) denotes whether the executed method is `run()` or `getFallback()`. For fallback invocations,
- `hystrix.command.fallback.cause` annotation is added that records the Exception that caused the fallback method to be invoked.

Note that the values of the annotations above are dependent on the hystrix-core version.
**pre 1.4.0** - Both are recorded on successful calls or fallbacks caused by execution exceptions. Any failures due to Hystrix internal checks (such as timeout, open circuits) will not record them.
**1.4.0-1.4.21** - `hystrix.command.execution` is recorded, but `hystrix.command.fallback.cause` is not.
**1.4.22+** - Both `hystrix.command.execution` and `hystrix.command.fallback.cause` are recorded.
### Issues with hystrix-core 1.5.4+

The `HystrixCommand` class got an additional anonymous inner class in 1.5.4 and unfortunately is compiled into `HystrixCommand$1`, which caused incorrect method to be traced for 1.5.4+. The fix is done by looping through N number of `HystrixCommand`'s anonymous inner classes (currently 3) and checking their enclosing method and return type to decide if it should be transformed and if so, which interceptor to inject.

@jiaqifeng Thanks for your work in the Hystrix plugin. It would have been impossible to do any further work without your code.
Could you review this pull request and let us know if anything needs fixing?


koo-taejin:
Can one of the admins verify this patch?

jiaqifeng:
@Xylus 
I just do a quick look at the code, it seems ok and no comments currently. I will look inside HystrixCommand deeper if I have time.

Xylus:
@jiaqifeng Thanks for taking a look.

Xylus:
@jiaqifeng Getting ready for 1.6.0 release, we'll merge this tomorrow if there's nothing seriously wrong.
Please do let us know if there's any immediate issue you'd like to address.
Thanks again!

Xylus:
@jiaqifeng Will go ahead and merge this for 1.6.0. Let us know if anything's not working if you test it out.

Issue #2106, opened 2016-09-26
Can we trace the standlone thrift RPC server?


duwupeng:
Hi All,

First of all, Thank you so much for the lovely tool. I have gone through it and found really interesting.

I have checked this by configuring agent on Tomcat and it's showing more than I expected.

Now, The question is :- Can we trace the  thrift RPC client and server? 

I look forward for your response.

Cheers,
duwupeng

2016-09-26 17:49:03 [DEBUG](c.n.p.b.i.s.ScopedInterceptor      :298) tryBefore() returns false: interceptorScopeTransaction: InterceptorScopeInvocation(ThriftClientScope)[depth=0], executionPoint: INTERNAL. Skip interceptor class com.navercorp.pinpoint.plugin.thrift.interceptor.tprotocol.client.TProtocolWriteFieldStopInterceptor
2016-09-26 17:49:03 [DEBUG](c.n.p.b.i.s.ScopedInterceptor      :298) tryAfter() returns false: interceptorScopeTransaction: InterceptorScopeInvocation(ThriftClientScope)[depth=0], executionPoint: INTERNAL. Skip interceptor class com.navercorp.pinpoint.plugin.thrift.interceptor.tprotocol.client.TProtocolWriteFieldStopInterceptor

duwupeng:
The thrift plugin has been enabled , while it can not trace the data. 
###########################################################
# Thrift                                                  #
###########################################################
# Profile Thrift
profiler.thrift.client=true
profiler.thrift.client.async=true
# Profile processor.
profiler.thrift.processor=true
profiler.thrift.processor.async=true
# Allow recording arguments.
profiler.thrift.service.args=true
# Allow recording result.
profiler.thrift.service.result=true

Xylus:
Hi duwupeng,
Could you give a short explanation on how you defined your processors? Are you using the default processor auto-generated from the IDL?

duwupeng:
Hell Xylus ,
Thanks for four reply。
 As you said , we defined the processors using the thrift Compiler (0.9.2) from IDL

Xylus:
Hmm..that's weird, will definitely need to look into it further.
Could you post the agent log please? It should be under '$AGENT_DIR/logs' directory.
Thanks!

duwupeng:
sure.  The processor and the agent log are as below. 
1. The agent log:

[...]

PR #2110, opened 2016-09-28
Adds support for TProtocolDecorators

Xylus:
`TProtocolDecorator` is added in 0.9.1.
When tracing Thrift processors, the agent marks whether to trace a request by setting an injected field to `TProtocol` instance when `ProcessFunction.process(int, TProtocol, TProtocol, I)` is invoked.
If the input protocol (2nd parameter) is an instance of `TProtocolDecorator`, the `ProcessFunction.process` interceptor marks it to the decorator, and not the actual `TProtocol` it is decorating.
This PR fixes this by retrieving the actual `TProtocol` instance before marking it.

for #2106


koo-taejin:
Can one of the admins verify this patch?

koo-taejin:
lgtm :)

PR #2121, opened 2016-10-06
added executor made of disruptor. #1804

koo-taejin:
There is no performance improvement in test results.
So delete all of the code that is related to disruptor.


koo-taejin:
ping @emeroad  :)

PR #2127, opened 2016-10-06
Changing hbase version in hbase/pom.xml

mvlbarcelos:
Changing hbase version because link of version1.2.1 doesn't exists more.


koo-taejin:
I have checked.
This site(Http://apache.mirrors.pair.com/) is not to provide for hbase 1.2.1 any more.
For the same reason, we are change hbase version Continuously.
After, I will solve this issue. 

Thank you for the contribution.

Issue #2142, opened 2016-10-12
Add component-scan interceptor for Spring beans plugin.


jaehong-kim:

jaehong-kim:
Spring XML's <context:component-scan> and @ComponentScan classes.

~~~
# filters
#    filter
#    filter OR filters
# filter
#    value
#    value AND filter
# value
#    token
#    token OR token
# token
#    profiler.spring.beans.n.scope= [component-scan | post-processor] default is component-scan.
#    profiler.spring.beans.n.base-packages= [package name, ...]
#    profiler.spring.beans.n.name.pattern= [regex pattern, regex:regex pattern, antstyle:antstyle pattern, ...]
#    profiler.spring.beans.n.class.pattern= [regex pattern, regex:regex pattern, antstyle:antstyle pattern, ...]
#    profiler.spring.beans.n.annotation= [annotation name, ...]
#
# Scope:
# component-scan: <context:component-scan ... /> or @ComponentScan
# post-processor: BeanPostProcessor - Slow!!!
#
# ANT Style pattern rules:
# ? - matches on character
# * - matches zero or more characters

# Examples
# profiler.spring.beans.1.name.scope=component-scan
# profiler.spring.beans.1.base-packages=com.foo, com.bar
# profiler.spring.beans.1.name.pattern=.*Foo, regex:.*Bar, antstyle:*Controller
# profiler.spring.beans.1.class.pattern=
# profiler.spring.beans.1.annotation=org.springframework.stereotype.Controller,org.springframework.stereotype.Service,org.springframework.stereotype.Repository
#
# profiler.spring.beans.2.name.scope=post-processor
# profiler.spring.beans.2.base-packages=com.foo
# profiler.spring.beans.2.name.pattern=
# profiler.spring.beans.2.class.pattern=antstyle:com.foo.repository.*Repository, antstyle:com.foo.Service.Main*
# profiler.spring.beans.2.annotation=
~~~

PR #2143, opened 2016-10-12
Spring beans transformer

jaehong-kim:
#2142


jaehong-kim:
ping @emeroad please.

Issue #2198, opened 2016-10-28
Fixed send retry policy in TCPDataSender


koo-taejin:
Retry policy in TCPDataSender is not working properly, so Fixed it.

PR #2199, opened 2016-10-28
Fixed send retry policy in TCPDataSender #2198

koo-taejin:
1. Fixed retry policy in TCPDataSender is not working properly.
2. Changed log level


koo-taejin:
ping @emeroad

PR #2263, opened 2016-11-18
#2264 fix ASMClass.addDelegatorMethod()

jaehong-kim:
#2264
* change INVOKESPECAIL class name.
* add local variables.


jaehong-kim:
#2264

PR #2289, opened 2016-11-28
add a hbase script

zcmjcd1:
It is useful for keeping the realtime agent list chart clean. When there are too many "not found", use it.


Xylus:
Hi @zcmjcd1,
The idea itself behind having a truncate script is good.
But selectively truncating tables just to clear agents from the realtime active thread chart will incur data loss in other parts of the web.
If you're trying to keep the agent list clean, there's admin APIs available that you can use (see [/removeAgentId.pinpoint](https://github.com/naver/pinpoint/blob/master/web/src/main/java/com/navercorp/pinpoint/web/controller/AdminController.java#L63))

Please let me know if I'm mistaken of your intentions (feel free to reopen the pull-request if this is the case).

Issue #2286, opened 2016-11-28
Add function of getting thread dump from agent


koo-taejin:
Add function of getting thread dump from agent

wingyiu:
doc plz?

PR #2344, opened 2016-12-14
Add function of getting thread dump from agent #2286

koo-taejin:
1. added function of getting activeThreadDump


koo-taejin:
@emeroad review please :)

PR #2348, opened 2016-12-15
Makes password for admin APIs optional

Xylus:
If `admin.password` is not set in *pinpoint-web.properties*, admin APIs do not require password parameter.


koo-taejin:
looks good to me :)

PR #2364, opened 2016-12-16
#2025 Allows send data that consist of Span and SpanChunk List

koo-taejin:
1. Allows send data that consist of Span and SpanChunk List in a specific situation.


koo-taejin:
@emeroad review please :)

PR #2369, opened 2016-12-16
Fix hbase pom

marcoblos:
When I use this `mvn -f hbase docker: build` command, to build the hbase project inside the pinpoint, I received this message 'returned a non-zero`.

When I looked at the pom.xml and I saw two different uses between the same variable.

So I changed `$ {HBASE_HOME}` to `$ HBASE_HOME` and everything works fine.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 /.project
 /.settings/
 /.idea
+/.vscode/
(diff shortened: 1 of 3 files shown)


koo-taejin:
thanks for contributing :)

PR #2376, opened 2016-12-19
Add script to install maven project and build all docker images

marcoblos:
It is necessary to run _docker-compose up_ without run command per command manually.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,0 +1,15 @@
+#!/usr/bin/env bash
+
+mvn install -Dmaven.test.skip=true
+
+echo '   >>>>> Create hbase image'
+mvn -f hbase package -Dmaven.test.skip=true
+mvn -f hbase docker:build
+
+echo '   >>>>> Create collector image'
+mvn -f collector package -Dmaven.test.skip=true
+mvn -f collector docker:build
+
+echo '   >>>>> Create web image'
+mvn -f web package -Dmaven.test.skip=true
+mvn -f web docker:build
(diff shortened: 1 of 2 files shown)


koo-taejin:
Hi marcoblos 

Building and Installing with scripts is a really good idea.

But there are a lot of things to consider.
1. whether maven executable file contains PATH environment variable
2. If not, whether MAVEN_HOME environment variable is set
3. If not, show enough error log 

In addition, 
It would be better to separate the docker image creation using additional parameter.
Because, creating docker image is very heavy, and can execute in specific situation. 

So, we can't merge this pull request. 

thanks.

marcoblos:
@ Koo-taejin Thanks for your reply. I will improve this feature.

In my personal repository I do it better, download maven, jdk and everything to make this work, but I think if I do all those things, the script would be very slow.

Do you have more ideias to improve this script?

I will think better about this script.

Thanks in advance.

koo-taejin:
@marcoblos 

I need more worries about related build a docker container.
I think it would be nice to include only the project build at now.

The following is a simple idea.
```
// usage
./install.sh -P{profile dev|release}

# check JAVA_6_HOME

# check JAVA_7_HOME

# check JAVA_8_HOME

# To check if Maven has been installed

# To check installed maven version higher than 3.2.x+

mvn clean install -Dmaven.test.skip=true -P{profile}
```

marcoblos:
Can you share your worries about build a docker images? There is a problem about pinpoint with docker?

Thanks @koo-taejin

koo-taejin:
If running this script in my development environment will fail.

It is not a good idea to make it executable only under certain circumstances where you can connect to the docker container with a generic name of install.sh.
If so, the default install.sh file is a simple project build that can be run in any environment, and building the docker image is optional.

In this case, you should try to figure out what the option name will do and if there are any additional options to extract.

PR #2410, opened 2016-12-30
 #2286 Add function of getting thread dump from agent

koo-taejin:
1. change response data
2. add test


koo-taejin:
@emeroad review please :)

PR #2427, opened 2017-01-04
Add support for Resin Plugin

bai-yang:
Dear all:
     Resin is a fast, ultra-reliable Java EE integrated solution for web applications spanning hundreds of servers. 

Resin Plugin Contribution
Link: /plugins/resin
Target: pinpoint-resin-plugin-1.6.1-SNAPSHOT.jar
Supported Version: resin-3.x.x and resin-4.x.x

ServiceTypes: RESIN=1111 ,RESIN_METHOD=1112
Annotations: 
Configurations:
profiler.resin.enable=
profiler.resin.bootstrap.main=
profiler.resin.excludeurl=

Example：

resin-0
resin-1


linuxmyjob:
hello,which version will support resin

Xylus:
@linuxmyjob it'll be merged into the current mater (1.6.1-SNAPSHOT).

linuxmyjob:
thanks,look forward to it !

Xylus:
Hi @bai-yang
Got it working on Resin 4.0.43 using the configuration suggested in #2450 and it seems to work great apart from UNDEFINED `doFilter` method in the call stack..
But I have the same problem running it on Resin 3.1.15 as mentioned in #2450 .
Is there a way to make this work with Resin 3.x?

Also, I've went over both Resin PRs and both look very promising. But we can only merge one of them as the plugins will conflict. 
So we'll have to choose.
The biggest (and probably the only) difference between them is the interception point (`WebAppFilterChain.doFilter` vs `ServletInvocation.service`).
While using the filter chain as the entry point works, I think `ServletInvocation.service` is more appropriate as the span's entry point. It also is more in line with other plugins such as Tomcat, JBoss, and Jetty.

Would it be possible for you to collapse on to #2450 and collaborate on the plugin? Your insights on the issue will be very valuable.
Thanks!

bai-yang:
@Xylus you are right，ServletInvocation.service is more appropriate as the span's entry point。I will have a try using the configuration suggested in #2450 to solve the problem running it on Resin 3.1.15。Thanks!

Xylus:
Thanks again @bai-yang 
Looking forward to your help! Really appreciate everything you've done.

linuxmyjob:
hello,1.7 support resin 3.x ?

PR #2450, opened 2017-01-06
Resin Plugin

huayaoyue6:
Hello ,I developed the resin plugin in November last year based on 1.5.2 branch. now I've upgraded it  to 1.6.0 .

##### Step One
This step mean to telling resin to load pinpoint plugins .
* put all pinpoint plugins in "plugin" folder which locate in "pinpoint-agent-1.6.0" folder into resin "webapp-jars" 
* Or add '\<library-loader path="/xxxx/pinpoint-agent-1.6.0/plugin"/>' in the "\<class-loader>" element,like this:
```xml
<web-app-default>
  <class-loader>
    <library-loader path="${resin.root}/webapp-jars"/>
    <library-loader path="${resin.root}/webapp.d/lib"/>
    <library-loader path="config:/webapp.d/lib"/>
    <library-loader path="/xxxx/pinpoint-agent-1.6.0/plugin"/>
  </class-loader>
</web-app-default>
```
* Other xml config Or folder different with config/folder mentioned above also maybe right If resin can load those pinpoint plugins correctly.

##### Step Two
* Add "\<jvm-arg>" element in the "\<server-default>" or others pleace can add jvm parameter config . config in xml like this:
```
 <jvm-arg>-javaagent:/xxxx/pinpoint-agent-1.6.0/pinpoint-bootstrap-1.6.0.jar</jvm-arg>
 <jvm-arg>-Dpinpoint.agentId=xxxx</jvm-arg>
 <jvm-arg>-Dpinpoint.applicationName=xxxx</jvm-arg>
```
##### How to build

The  resin-4.0.35.jar  download from http://caucho.com/ and the jar already added in  "plugins/resin/lib/" folder.
Another one "resin.jar" is in the version 3.1.15,also downloaded from [Offical website](http://caucho.com/products/resin/download/3-1/gpl) .
Those dependency can install by hand us :
```shell
mvn install:install-file -Dfile=resin.jar -DgroupId=com.caucho -DartifactId=resin -Dversion=3.1.15 -Dpackaging=jar
```
Because of the commit is too large with those two jar.You can recommit and remove those jar . 

##### Code
You can rewrite all code or whatever. During the plugin developing  time , I used tomcat plugin for reference .


Xylus:
Hi @huayaoyue6 
I've tested this on Resin 4.x and it works out great.

Is there a way to run this on Resin 3.x? I've tried running it on Resin 3.1.15 but it blows up during startup.
(I've made changes for the plugin to not have compile/run time dependencies to Resin 4.x to test this out which I'll suggest in the PR review soon.)
Is there any thing else I need to configure to run it on Resin 3.x?

huayaoyue6:
OK, I will try it . I develeoped and testd on Resin 4.0.35. Did not test it on Resin3.X

huayaoyue6:
OK, I will try it . I develeoped  and testd  the plugin on Resin 4.0.35. Hadn't  tested  it on Resin3.X

commons/src/main/java/com/navercorp/pinpoint/common/util/ClassLoaderUtils.java
```diff
@@ -34,15 +34,29 @@ public ClassLoader getClassLoader() {
...
+	            EXT_CLASS_LOADER = SYSTEM_CLASS_LOADER.getParent();
+	        } else {
+	            EXT_CLASS_LOADER = null;
+	        }
+	        if (EXT_CLASS_LOADER != null) {
+	            BOOT_CLASS_LOADER = EXT_CLASS_LOADER.getParent();
+	        } else {
+	            BOOT_CLASS_LOADER = null;
+	        }
         }
```
Xylus, line 46: Is this change still necessary? I tried testing it out without this change on Resin 4.0.49 (Java 1.8) and it seemed to work just fine. Could you test it out without the change and see if it works?
I ask because it's best not to introduce any Resin specific code into the common module.

plugins/resin/pom.xml
```diff
@@ -0,0 +1,37 @@
...
+		<dependency>
+			<groupId>javax.servlet</groupId>
+			<artifactId>javax.servlet-api</artifactId>
+			<scope>compile</scope>
+		</dependency>
+
+		<dependency>
+			<groupId>com.caucho</groupId>
+			<artifactId>resin</artifactId>
+			<version>4.0.35</version>
```
Xylus: Introducing a dependency that is not readily available will make installing pinpoint more complicated.
I'd suggest using the one available in maven central which seems to be *3.0.9*.

However, making this change will break some code that is compile-time dependent to 4.0.35. I'll suggest further changes down below.

agent/src/main/resources-release/pinpoint.config
```diff
@@ -574,4 +574,19 @@ profiler.activemq.client.destination.exclude=
...
+# default enable resin plugin
+profiler.resin.enable=true
+# if empty , default value is : com.caucho.server.resin.Resin
+profiler.resin.bootstrap.main=
+# trace param in request  ,default value is true
+profiler.resin.tracerequestparam=true
+# excudeurl eg: filter static resources : /**/*.jpg,/**/*.png,/**/*.css,/**/*.js
+profiler.resin.excludeurl=
+# trace cookies in request ,default value is true
+profiler.resin.tracecookies=true
```
Xylus, line 631: Need these options in `agent/src/main/resources-local/pinpoint.config` and `agent/src/main/resources/pinpoint-real-env-lowoverhead-sample.config` as well.

plugins/resin/src/main/java/com/navercorp/pinpoint/plugin/resin/ResinPlugin.java
```diff
@@ -0,0 +1,216 @@
...
+//                InstrumentClass target = instrumentor.getInstrumentClass(classLoader, className, classfileBuffer);
+//                if (target != null) {
+//                    target.weave("com.navercorp.pinpoint.plugin.resin.aspect.RequestFacadeAspect");
+//                    return target.toBytecode();
+//                }
+//
+//                return null;
+//            }
+//        });
+//    }
```
Xylus: Could probably delete this :)

plugins/resin/src/main/java/com/navercorp/pinpoint/plugin/resin/aspect/RequestFacadeAspect.java
```diff
@@ -0,0 +1,67 @@
...
+import com.navercorp.pinpoint.bootstrap.instrument.aspect.PointCut;
+
+import java.util.Enumeration;
+
+/**
+ * filtering pinpoint header
+ * @author emeroad
+ */
+@Aspect
+public abstract class RequestFacadeAspect {
```
Xylus: Is this class necessary? It doesn't seem to be used.

plugins/resin/src/main/java/com/navercorp/pinpoint/plugin/resin/interceptor/ServletInvocationInterceptor.java
```diff
@@ -0,0 +1,467 @@
...
+		if (getTraceMetadata(request) == null) {
+			return false;
+		}
+
+		return getAsyncMetadata(request);
+	}
+
+	private boolean getAsyncMetadata(final HttpServletRequest request) {
+		if(request instanceof com.caucho.server.webapp.AsyncRequest){
+			HttpServletRequest servletRequest = ((com.caucho.server.webapp.AsyncRequest) request).getRequest();
```
Xylus: `com.caucho.server.webapp.AsyncRequest` brings in dependency not available in Resin 3.x, which will throw out exceptions when running the agent on Resin 3.x.

You could work around this issue by making a "Getter" class for `HttpServletRequest` and add it to  `com.caucho.server.http.CauchoRequestWrapper`'s transformer like the following :
(`CauchoRequestWrapper` seems to be `AsyncRequest`'s parent class)
```
public interface HttpServletRequestGetter {
    HttpServletRequest _$PINPOINT$_getRequest();
}
```
ResinPlugin.java
```
transformTemplate.transform("com.caucho.server.http.CauchoRequestWrapper", new TransformCallback() {

    @Override
    public byte[] doInTransform(Instrumentor instrumentor, ClassLoader classLoader, String className, Class<?> classBeingRedefined, ProtectionDomain protectionDomain, byte[] classfileBuffer) throws InstrumentException {
        InstrumentClass target = instrumentor.getInstrumentClass(classLoader, className, classfileBuffer);

        target.addGetter("com.navercorp.pinpoint.plugin.resin.HttpServletRequestGetter", "_request");
        return target.toBytecode();
    }
});
```
ServletInvocationInterceptor.java
```
if (request instanceof HttpServletRequestGetter) {
    HttpServletRequest servletRequest = ((HttpServletRequestGetter) request)._$PINPOINT$_getRequest();
```

With these changes, running on Resin 3.x will simply skip all this and won't throw an exception.

plugins/resin/src/main/java/com/navercorp/pinpoint/plugin/resin/interceptor/ServletInvocationInterceptor.java
```diff
@@ -0,0 +1,467 @@
...
+			return false;
+		}
+
+		return ((AsyncAccessor) request)._$PINPOINT$_isAsync();
+	}
+
+	private Trace getTraceMetadata(final HttpServletRequest request) {
+		if(request instanceof com.caucho.server.webapp.AsyncRequest){
+			
+			HttpServletRequest servletRequest = ((com.caucho.server.webapp.AsyncRequest) request).getRequest();
```
Xylus: Same as above.

plugins/resin/src/main/java/com/navercorp/pinpoint/plugin/resin/interceptor/WebAppInterceptor.java
```diff
@@ -0,0 +1,187 @@
...
+			return;
+
+		_jarList.add(jarEntry);
+
+		JarMap _pathMap = new JarMap();
+
+		if (_pathMap != null) {
+			_pathMap.scan(jar, jarEntry);
+		}
+	}
```
Xylus: You don't really need to create new `JarPath`s, `JarEntry`s, and `JarMap`s here and read in the actual file.
I think simply using the String representation of Path should be enough. Plus, by not using `JarEntry`, `JarMap` here, you can remove these (Caucho-copied) classes safely.

plugins/resin/src/main/java/com/navercorp/pinpoint/plugin/resin/ResinPlugin.java
```diff
@@ -0,0 +1,216 @@
...
+		transformTemplate.transform("com.caucho.server.http.HttpRequest", new TransformCallback() {
+
+			@Override
+			public byte[] doInTransform(Instrumentor instrumentor, ClassLoader classLoader, String className, Class<?> classBeingRedefined, ProtectionDomain protectionDomain, byte[] classfileBuffer) throws InstrumentException {
+				InstrumentClass target = instrumentor.getInstrumentClass(classLoader, className, classfileBuffer);
+				
+				// clear request.
+				InstrumentMethod finishInvocationMethodEditor = target.getDeclaredMethod("finishRequest");
+				if (finishInvocationMethodEditor != null) {
+					finishInvocationMethodEditor.addInterceptor("com.navercorp.pinpoint.plugin.resin.interceptor.HttpRequestInterceptor");
```
Xylus: This interceptor could be attached to `com.caucho.server.http.HttpServletRequestImpl.finishRequest()` method.
Please see the [comment](#diff-bd112311a7bbcb0d4b6e981a27a7e283) made for `HttpRequestInterceptor` for more details.

plugins/resin/src/main/java/com/navercorp/pinpoint/plugin/resin/interceptor/HttpRequestInterceptor.java
```diff
@@ -0,0 +1,73 @@
...
+	public HttpRequestInterceptor(InstrumentMethod targetMethod) {
+		this.targetMethod = targetMethod;
+	}
+
+	@Override
+	public void before(Object target, Object[] args) {
+		logger.beforeInterceptor(target, target.getClass().getName(), targetMethod.getName(), "", args);
+		try {
+			com.caucho.server.http.HttpRequest httpRequest = (com.caucho.server.http.HttpRequest) target;
+			HttpServletRequest httpServletRequest = (HttpServletRequest)httpRequest.getRequestFacade();
```
Xylus: `com.caucho.server.http.HttpRequest` (and it's parent class `AbstractHttpRequest`) in Resin 3.x does not seem to have `getRequestFacade()` method and would throw exceptions when running the agent on Resin 3.x.

You could attach this interceptor to hook on `com.caucho.server.http.HttpServletRequestImpl.finishRequest()`. The `target` object would then be an instance of `AsyncAccessor` and `TraceAccessor` and you'll be able to use the methods directly like so
```
if(target != null){
    if (target instanceof AsyncAccessor) {
        // reset
        ((AsyncAccessor) target)._$PINPOINT$_setAsync(Boolean.FALSE);
    }
    if (target instanceof TraceAccessor) {
        ...
```

huayaoyue6:
@Xylus Hello there ,I have already update the resin code . I find the dependency "resin.jar" downloaded from maven repository and the coordinate copied from (http://mvnrepository.com)  is not as same content as the one which download from [Offical website](http://caucho.com/products/resin/download/3-1/gpl) . So I just add the offical jar lib named "resin.jar " into "plugin/resin/lib/" . :smile:

huayaoyue6:
@Xylus I've edit the top or the first comment.Add pinpoint's agent config in resin.You maybe need to read the new content.

plugins/resin/pom.xml
```diff
@@ -24,13 +24,13 @@
 			<artifactId>javax.servlet-api</artifactId>
 			<scope>compile</scope>
```
Xylus: This should be `provided`.

plugins/resin/pom.xml
```diff
@@ -24,13 +24,13 @@
...
 		</dependency>
-
 		<dependency>
-			<groupId>com.caucho</groupId>
-			<artifactId>resin</artifactId>
-			<version>4.0.35</version>
-			<scope>provided</scope>
+		    <groupId>com.caucho</groupId>
+		    <artifactId>resin</artifactId>
+		    <version>3.1.15</version>
```
Xylus: Let's use the version available from maven central - `3.0.9`.
(Pinpoint users shouldn't have to install resin into their local repository)

plugins/resin/pom.xml
```diff
@@ -24,13 +24,13 @@
...
+		    <scope>compile</scope>
```
Xylus: This should be `provided` as well.

plugins/resin/src/main/java/com/navercorp/pinpoint/plugin/resin/ResinPlugin.java
```diff
@@ -149,53 +130,46 @@ private void addServletInvocation() {
 				if (serviceMethodEditorBuilder != null) {
 					serviceMethodEditorBuilder.addInterceptor("com.navercorp.pinpoint.plugin.resin.interceptor.ServletInvocationInterceptor");
```
Xylus: You could change this to
```
serviceMethodEditorBuilder.addScopedInterceptor("com.navercorp.pinpoint.plugin.resin.interceptor.ServletInvocationInterceptor", $SCOPE_NAME, ExecutionPolicy.BOUNDARY);
```
so that you don't have to do the `checkRequestDispatcherImpl()` check inside `ServletInvocationInterceptor`.

Scoped interceptors with `ExecutionPolicy.BOUNDARY` will only run when they're not inside the same scope (identified by $SCOPE_NAME above). So in this case, `ServletInvocationInterceptor` will only run once at the beginning of a request, and won't run when it's invoked by `RequestDispatcherImpl` as it's inside the scope already.

plugins/resin/src/main/java/com/navercorp/pinpoint/plugin/resin/interceptor/ServletInvocationInterceptor.java
```diff
@@ -106,23 +119,21 @@ public void before(Object target, Object[] args) {
 			}
 		}
 	}
-	
-	
-	public boolean checkRequestDispatcherImpl(){
+
+	public boolean checkRequestDispatcherImpl() {
```
Xylus: Retrieving stack traces every request will be a big hit in performance.
This method won't be needed after you add `ServletInvocationInterceptor` as a scoped interceptor (see the comments in `ResinPlugin` for details on this).

plugins/resin/src/main/java/com/navercorp/pinpoint/plugin/resin/interceptor/WebAppInterceptor.java
```diff
@@ -1,38 +1,27 @@
...
-
-
 
 /**
  * 
- * @author huangpengjie@fang.com 
+ * @author huangpengjie@fang.com
  *
  */
 public class WebAppInterceptor implements AroundInterceptor {
```
Xylus, line 23: Let's try to get rid of the compile-time dependency to `com.caucho.server.webapp.WebApp` as this class is not available in *3.0.9*.

You can do this by changing all `WebApp` to `ServletContext` as `WebApp` is a child of `ServletContext`.

plugins/resin/src/main/java/com/navercorp/pinpoint/plugin/resin/interceptor/WebAppInterceptor.java
```diff
@@ -61,127 +50,75 @@ public void after(Object target, Object[] args, Object result, Throwable throwab
...
 
 	private String extractContextKey(WebApp webapp) {
-		return webapp.getContextPath();
+		String context =  webapp.getContextPath();
+		return (context == null || context.length() == 0) ? "/ROOT": context;
 	}
-	
+
 	private List<String> extractLibJars(WebApp webapp) {
 		EnvironmentClassLoader classLoader = webapp.getEnvironmentClassLoader();
```
Xylus: If you change `WebApp` to `ServletContext`, this part will break as `ServletContext` does not have `getEnvironmentClassLoader()` method.

You can fix this by doing something like:
```
private List<String> extractLibJars(ServletContext webapp) {
    ClassLoader classLoader = webapp.getClassLoader();
    if (classLoader instanceof URLClassLoader) {
        URL[] urls = ((URLClassLoader) classLoader).getURLs();
        return extractLibJarNamesFromURLs(urls);
    } else {
        return Collections.emptyList();
    }
}
```

plugins/resin/src/main/java/com/navercorp/pinpoint/plugin/resin/interceptor/WebAppInterceptor.java
```diff
@@ -61,127 +50,75 @@ public void after(Object target, Object[] args, Object result, Throwable throwab
...
+					logger.debug("{}  jars : {}", contextKey, Arrays.toString(loadedJarNames.toArray()));
 				}
-				dispatchLibJars(contextKey, loadedJarNames,webapp);
+				dispatchLibJars(contextKey, loadedJarNames, webapp);
 			} catch (Exception e) {
-					logger.warn(e.getMessage(), e);
+				logger.warn(e.getMessage(), e);
 			}
 		} else {
 			logger.warn("Webapp loader is not an instance of org.apache.catalina.loader.WebappLoader. Found [{}]", target.getClass().toString());
```
Xylus: The log message seems like it's for Tomcat :)

huayaoyue6:
I've done resin plugin completely . I think it's hard to find a suitable demo to test  a server's  basic features.
And maybe [spring-mvc-showcase](https://github.com/spring-projects/spring-mvc-showcase) is a useful demo for develop web server plugin . 
@Xylus

Xylus:
@huayaoyue6 Everything looks good, thanks for your contribution!

zvrr:
finally, whether can work in resin-3.1.15?

Issue #2496, opened 2017-01-17
Add datasource metric collection


koo-taejin:
Gather information from a data source for per-agent.

* ServiceType
* PoolName
* JDBCURL
* ActiveConnectionCount
* MaxConnectionCount

PR #2502, opened 2017-01-18

koo-taejin:
1. Create classes related to support plugin monitor.


koo-taejin:
@emeroad review please :)

PR #2515, opened 2017-01-20

koo-taejin:
1. Moving class's modules
2. Moving class's package


koo-taejin:
@emeroad review please :)

PR #2537, opened 2017-01-26

koo-taejin:
1. add handler for datasource


koo-taejin:
@Xylus review please :)

PR #2547, opened 2017-02-01

koo-taejin:
add controller for to read datasource metric

### DIFF
--- a/web/src/main/java/com/navercorp/pinpoint/web/mapper/stat/SampledDataSourceResultExtractor.java
+++ b/web/src/main/java/com/navercorp/pinpoint/web/mapper/stat/SampledDataSourceResultExtractor.java
@@ -1,0 +1,117 @@
+/*
+ * Copyright 2017 NAVER Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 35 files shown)


web/src/main/java/com/navercorp/pinpoint/web/mapper/stat/SampledDataSourceResultExtractor.java
```diff
@@ -0,0 +1,117 @@
...
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * @author Taejin Koo
+ */
+public class SampledDataSourceResultExtractor extends SampledAgentStatResultExtractor {
```
Xylus: Could this be
```
public class SampledDataSourceResultExtractor implements ResultsExtractor<List<List<SampledDataSource>> {
```
and the `SampledDataSourceList` class might not be needed by using `List<SampledDataSource>` instead in other classes that reference it?
Might be a bit simpler this way, not sure if it's workable though.

koo-taejin:
@Xylus Thank you for your review.
It was a my mistake. I changed it. :)

Xylus:
:+1:

PR #2674, opened 2017-03-07

koo-taejin:
Datasource metric collection support in hikaricp.


plugins/hikaricp/src/main/java/com/navercorp/pinpoint/plugin/hikaricp/HikariCpPlugin.java
```diff
@@ -72,26 +62,88 @@ private void addBasicDataSourceTransformer() {
...
+                    addDataSourceMonitorInterceptor(target, constructor);
+                    return target.toBytecode();
+                }
+            }
+            return target.toBytecode();
+        }
+
+        private boolean isAvailableDataSourceMonitor(InstrumentClass target) {
+            InstrumentMethod getActiveConnections = target.getDeclaredMethod("getActiveConnections");
+            if (getActiveConnections == null || !int.class.getName().equals(getActiveConnections.getReturnType())) {
```
jaehong-kim: It is better to consider the case where the result value of getActiveConnections.getReturnType () is null.
JavassistMethod.java
~~~
    @Override
    public String getReturnType() {
        if (behavior instanceof CtMethod) {
            try {
                return ((CtMethod) behavior).getReturnType().getName();
            } catch (NotFoundException e) {
                return null;
            }
        }

        return null;
    }
~~~

plugins/hikaricp/src/main/java/com/navercorp/pinpoint/plugin/hikaricp/interceptor/DataSourceConstructorInterceptor.java
```diff
@@ -0,0 +1,94 @@
...
+        if (args.length >= 1) {
+            try {
+                String jdbcUrl = getJdbcUrl(args[0]);
+                HikariCpDataSourceMonitor dataSourceMonitor = new HikariCpDataSourceMonitor(target, jdbcUrl);
+                dataSourceMonitorRegistry.register(dataSourceMonitor);
+
+                if (target instanceof DataSourceMonitorAccessor) {
+                    ((DataSourceMonitorAccessor) target)._$PINPOINT$_setDataSourceMonitor(dataSourceMonitor);
+                }
+            } catch (NoSuchMethodException e) {
```
jaehong-kim: It would be nice to consider an IllegalArgumentException.

koo-taejin: I fixed it. thanks :)

PR #2728, opened 2017-03-16
fixed typo in NOTICE.

fuzzythecat:


Xylus:
Good catch! Thank you for this :+1:

PR #2752, opened 2017-03-28
[#2751] refactoring cluster code for commonization 

minwoo-jung:
please review @koo-taejin


minwoo-jung:
I will merge after 1.6.1 release

PR #2766, opened 2017-04-03
ability to send all traffic in TCP from agent to collector

### LINKED ISSUE #2666: all agent traffic in only TCP
a configurable option in pinpoint agent config so all traffic from agent is in only TCP., it opens up options for further enhancements., ex., more load balancer options for collector

teambullseye:
resolve #2666
use the option in pinpoint.config to switch to all TCP or both TCP and UDP
when the option is set to true, all span and stat data will be sent in TCP
# Set all connections to TCP
profiler.all.tcp.enable=false


koo-taejin:
Hi teambullseye 

It would be better to let users choose TCP and UDP for each of their settings so they can decide for their own environment.

The configuration may be in the following form.
```
profiler.collector.span.transport.type = TCP / UDP (Default UDP)
```

I suggest you to fix it. 

thanks :)

teambullseye:
sure, I will make those changes and make a PR again

PR #2829, opened 2017-04-18
Modifies Thrift ITs to test against libthrift 0.10.x

### LINKED ISSUE #2522: Fix Thrift plugin integration tests to run on 0.10.x
#### Description
~~Thrift 0.10.0 is out with some internal changes that is currently not compatible with the thrift plugin.
Thrift plugin needs to be updated to support this.~~
Thrift 0.10.0 introduces breaking changes to auto-generated `AsyncIface` methods, which requires regenerating test service to run on libthrift 0.10.0.
The internals have not been changed as far as tracing with Thrift plugin is concerned. Only integration tests need fixing.

Xylus:
libthrift 0.9.x -> 0.10.x introduces breaking changes to auto-generated asynchronous client code (specifically the callback signature).

Now that the agent's libthrift version is upgraded to 0.10.0, we can now generate test service used by the ITs using 0.10.0 compiler and run ITs for libthrift 0.10.0.
Unfortunately, this means any ITs that tested asynchronous client invocations cannot be tested for 0.9.x (since again, incompatibility in generated code).

resolves #2522


Xylus:
Closing without merging - testing for 0.9.x takes priority over 0.10.x for regression.

PR #2852, opened 2017-04-24
Tcpconfigchanges

### LINKED ISSUE #2666: all agent traffic in only TCP

teambullseye:
resolve #2666
use the options in pinpoint.config to set transport type of span and stat 
default is UDP
# Set connection types for span and stat data. accepts UDP or TCP
profiler.collector.span.transport.type=UDP
profiler.collector.stat.transport.type=UDP

### DIFF
--- a/agent/pom.xml
+++ b/agent/pom.xml
@@ -330,7 +330,6 @@
                     </descriptors>
                     <outputDirectory>${project.build.directory}</outputDirectory>
                     <appendAssemblyId>false</appendAssemblyId>
-			<tarLongFileMode>posix</tarLongFileMode>
                 </configuration>
                 <executions>
                     <execution>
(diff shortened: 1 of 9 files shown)


koo-taejin:
Hi teambullseye 

I said following 
```
Each data type(Span, Stat, ...) must be on a different port.
Looking at the collector settings, we specify the threadPool for each data type because the importance and the incoming amount are different depending on the data type.
In addition, The same data type in the collector should be able to share the ThreadPool and set it collectively, regardless of trasnport.

In addition, It is a no good idea for a TCPHandler to accept all data types.
Because there are more parts to debug when a problem occurs
```

It does not seem to be working.

please check again. :)

smamidala1:
didnt realize you want them on different ports, will make those changes. thanks

Issue #2862, opened 2017-04-26
Workaround java level Deadlock when initializing annotation


emeroad:
#### Description
Class.getAnnotation() of JDK 6, 7 can cause java level deallock 

Related issue:
 - https://ecosystem.atlassian.net/browse/AO-283
 - https://bugs.openjdk.java.net/browse/JDK-7171421
 - https://issues.apache.org/jira/browse/HIVE-7409
 - http://stap.sjtu.edu.cn/index.php?title=Jdk7122142

#### Environment
Affects Version/s: 1.0.x, 1.1.x, 1.5.x, 1.6.1-

#### Additional Info
```
[Thread Dump] 
Found one Java-level deadlock:
=============================
"TP-Processor4096":
waiting to lock monitor 0x00007fd0e80094b0 (object 0x0000000761528120, a java.lang.Class),
which is held by "TP-Processor15"
"TP-Processor15":
waiting to lock monitor 0x00007fd0e8009a98 (object 0x000000079e0a1cb0, a org.apache.catalina.loader.WebappClassLoader),
which is held by "TP-Processor3"
"TP-Processor3":
waiting to lock monitor 0x00007fd0e80094b0 (object 0x0000000761528120, a java.lang.Class),
which is held by "TP-Processor15"

Java stack information for the threads listed above:
===================================================

"TP-Processor4096":
at sun.reflect.annotation.AnnotationType.getInstance(AnnotationType.java:63)
- waiting to lock <0x0000000761528120> (a java.lang.Class for sun.reflect.annotation.AnnotationType)
at sun.reflect.annotation.AnnotationParser.parseAnnotation(AnnotationParser.java:202)
at sun.reflect.annotation.AnnotationParser.parseAnnotations2(AnnotationParser.java:69)
at sun.reflect.annotation.AnnotationParser.parseAnnotations(AnnotationParser.java:52)
at java.lang.reflect.Method.declaredAnnotations(Method.java:693)
- locked <0x000000079e6c5678> (a java.lang.reflect.Method)
at java.lang.reflect.Method.getAnnotation(Method.java:679)
at com.nhn.lucy.web.interceptor.PropertyInterceptor.intercept(PropertyInterceptor.java:39)
...

"TP-Processor15":
at java.lang.Class.getDeclaredMethods0(Native Method)
at java.lang.Class.privateGetDeclaredMethods(Class.java:2436)
at java.lang.Class.getDeclaredMethods(Class.java:1793)
[...]

PR #2877, opened 2017-04-27
[#2862] remove annotation

koo-taejin:
dbcp, dbcp2, cxf


koo-taejin:
@minwoo-jung review please :)

Issue #2905, opened 2017-05-09
DubboProviderInterceptor has a bug,  it has not send span to collector


niyuelin1990:
2017-05-09 16:27:25 [WARN ](.n.p.p.d.i.DubboProviderInterceptor:403) BEFORE. Caused:already Trace Object exist.
com.navercorp.pinpoint.exception.PinpointException: already Trace Object exist.
	at com.navercorp.pinpoint.profiler.context.ThreadLocalTraceFactory.checkBeforeTraceObject(ThreadLocalTraceFactory.java:118)
	at com.navercorp.pinpoint.profiler.context.ThreadLocalTraceFactory.newTraceObject(ThreadLocalTraceFactory.java:128)
	at com.navercorp.pinpoint.profiler.context.active.ActiveTraceFactory.newTraceObject(ActiveTraceFactory.java:102)
	at com.navercorp.pinpoint.profiler.context.DefaultTraceContext.newTraceObject(DefaultTraceContext.java:162)
	at com.navercorp.pinpoint.plugin.dubbo.interceptor.DubboProviderInterceptor.createTrace(DubboProviderInterceptor.java:41)
	at com.navercorp.pinpoint.bootstrap.interceptor.SpanSimpleAroundInterceptor.before(SpanSimpleAroundInterceptor.java:51)
	at com.navercorp.pinpoint.bootstrap.interceptor.ExceptionHandleAroundInterceptor.before(ExceptionHandleAroundInterceptor.java:35)
	at com.alibaba.dubbo.rpc.proxy.AbstractProxyInvoker.invoke(AbstractProxyInvoker.java)
	at com.alibaba.dubbo.rpc.protocol.InvokerWrapper.invoke(InvokerWrapper.java:53)
	at com.alibaba.dubbo.rpc.filter.ExceptionFilter.invoke(ExceptionFilter.java:64)
	at com.alibaba.dubbo.rpc.protocol.ProtocolFilterWrapper$1.invoke(ProtocolFilterWrapper.java:91)
	at com.alibaba.dubbo.rpc.filter.TimeoutFilter.invoke(TimeoutFilter.java:42)
	at com.alibaba.dubbo.rpc.protocol.ProtocolFilterWrapper$1.invoke(ProtocolFilterWrapper.java:91)
	at com.alibaba.dubbo.monitor.support.MonitorFilter.invoke(MonitorFilter.java:75)
[...]

PR #2909, opened 2017-05-11
[#2905] DubboProviderInterceptor has a bug, it has not send span to c…

majinkai:


plugins/dubbo/src/main/java/com/navercorp/pinpoint/plugin/dubbo/interceptor/DubboConsumerInterceptor.java
```diff
@@ -94,7 +94,7 @@ private Trace getTrace(Object target) {
             return null;
         }
 
-        return traceContext.currentTraceObject();
+        return traceContext.currentRawTraceObject();
```
Xylus, line 97: I think this is fine for the `before()` method, but `after()` method should still use `traceContext.currentTraceObject()` or have a `trace.canSampled()` check inside it as well - like [here](https://github.com/naver/pinpoint/pull/3003/commits/c38ab80657b3a8bf47d92cc5e521a61769a16ec7#diff-be9037c2845c090663a9717aebf6c466)?

jiaqifeng: agree

Issue #2654, opened 2017-03-01
want to contribute to pinpoint community where see whether we can modify Http Interceptor to read the Http response code to distinguish between pass/fail requests


dinesh4747:
HI Pp Team -

Could you help me in brain storming on How to go about developing a capability to read the HTTP response code like 4xx/5xx from the logs and also which would leverage to distinguish between pass/fail requests 

Also make the scatter graph to display in red for all the failed requests 

Kindly let me know the appropriate class file, HTTPClient plugin where response code is thrown 

See whether any possibiity of modifying the code to ensure pinpoint collector capture the 5xx/4xx and treat them as failure and subsequently plot it in the server map

Xylus:
Hi @dinesh4747 
We might have a PR incoming regarding this exact issue (https://groups.google.com/forum/#!topic/pinpoint_user/miqxcXd31AQ).
If you can wait a bit, it'd be best to see how that goes before we start having multiple competing PRs.
Thanks.

PR #2919, opened 2017-05-12
tomcat profiler fail response codes

teambullseye:
resolves issue #2654 
this adds capability to tomcat plugin to read the HTTP response codes like 4xx/5xx and distinguish between pass/fail requests, also makes scatter chart to display in red for all the failed requests
and call stack shows the response code in red.
this option can be enabled or disabled and also success codes can be specified in pinpoint.config

### DIFF
--- a/web/src/main/java/com/navercorp/pinpoint/web/service/SpanServiceImpl.java
+++ b/web/src/main/java/com/navercorp/pinpoint/web/service/SpanServiceImpl.java
@@ -355,6 +355,11 @@ public class SpanServiceImpl implements SpanService {
 
     private StringMetaDataBo selectStringMetaData(String agentId, int cacheId, long agentStartTime) {
         final List<StringMetaDataBo> metaDataList = stringMetaDataDao.getStringMetaData(agentId, agentStartTime, cacheId);
+		if (cacheId == 404) {
+			StringMetaDataBo stringMetaDataBo = new StringMetaDataBo(agentId, agentStartTime, cacheId);
+			stringMetaDataBo.setStringValue("HTTP Failed response code");
+			return stringMetaDataBo;
+		}
         if (CollectionUtils.isEmpty(metaDataList)) {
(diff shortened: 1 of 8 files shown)


dinesh4747:
Hope the respective PR would be merged in the next release 1.7.0

web/src/main/java/com/navercorp/pinpoint/web/service/SpanServiceImpl.java
```diff
@@ -355,6 +355,11 @@ private void transitionException(List<SpanAlign> spanAlignList) {
 
     private StringMetaDataBo selectStringMetaData(String agentId, int cacheId, long agentStartTime) {
         final List<StringMetaDataBo> metaDataList = stringMetaDataDao.getStringMetaData(agentId, agentStartTime, cacheId);
+		if (cacheId == 404) {
```
emeroad: The cacheId is dynamically generated. Developers should not assign the cacheId arbitrarily.

dinesh4747:
Any plans of integrating this PR into upcoming versions ?

web/src/main/java/com/navercorp/pinpoint/web/service/SpanServiceImpl.java
teambullseye: Hi emeroad so we came up with two options we think will make this compatible 

one we could change the random range to be 500 to 5000 rather than 0 to 5000
within com.navercorp.pinpoint.common.server.bo on line 120 exceptionInfo.setIntValue(RandomUtils.nextInt(1000, 5000));

or we could simply change our 404 int on the identifier we are using to 5001 since that is outside the bounds of the current randomly assigned int

We really think having a set range that can be used for custom identifiers is a good approach for this and possible future customized exception options within pinpoint...can you let us know if either of these approaches will work for you all and if so which one?

dinesh4747:
@emeroad @Xylus - Is there any pipeline in place for accommodating this PR in the immediate releases? , This PR would absolutely going to add potential value in finding the failure request based on the HTTP response code filtering

emeroad:
Implemented in #3399

PR #2153, opened 2016-10-14
Websphere plugin

sjmittal:
the pull request is for websphere plugin.
there is also an experimental plugin axis2-jaxws to intercept webservices requests using axis2-jaxws.

PR #2933, opened 2017-05-18
Add websphere plugin & rebase. 

jaehong-kim:
https://github.com/naver/pinpoint/pull/2153


sjmittal:
Looks good to me. We will again check if resloveClass is needed or not after this is merged, and if we find that same is needed for websphere containers then we will issue a separate PR to discuss the same, as that is indeed a separate issue.

Thanks!

Issue #2970, opened 2017-05-30
aggregate agent stat data by using flink


minwoo-jung:
We can see agent stat data(cpu, memory, Transaction/thread count etc.) in inspector view.
But we can't see application-level (whole of agents) stat data.
**We will show application stat by using flink streaming. flink server aggregates agent stat data in real time.**

When an application has multiple agents, you currently have to check the inspector view of each agents separately in order to find abnormal agent(ex. using max memory).
With application stat data, you would be able to find any abnormal agent very easily.

PR #2971, opened 2017-05-30
[#2970] aggregate agent stat data by using flink

minwoo-jung:


jiaqifeng:
@minwoo-jung 
Woo, it is a BIG commit. I just wonder what function is added. Could you give a simple explain about what kind of agent stat is aggregated? Or any screenshot of new web ui is added which could explain more clearly?
Thanks.

minwoo-jung:
@jiaqifeng
Sorry. the answer is too late.  please ref [application inspector.](https://github.com/naver/pinpoint/blob/master/doc/application-inspector.md)

PR #3003, opened 2017-06-05
Fixes Dubbo consumer interceptor to propagate disable flags

### LINKED ISSUE #3002: Dubbo consumer interceptor does not propagate disable sampling flag to remote nodes
#### Description
`DubboConsumerInterceptor` checks for the current trace using `TraceContext.currentTraceObject()`. This API returns the current trace object only if the current trace should be sampled. It returns null when no trace object exists **or** when the current trace shouldn't be sampled. So even if a disabled trace object exists, it doesn't propagate this information over RPC.
We should instead use `TraceContext.currentRawTraceObject()` and check for sampling with the returned trace object.

issue found from #2989

Xylus:
resolves #3002


Xylus:
will instead fix via #2909

PR #3026, opened 2017-06-12
modify modify map caller and callee

ixufeng:


emeroad:
Thank you for your contributions

Issue #3114, opened 2017-07-06
Add alarm for datasource connection usage rate


koo-taejin:
Add alarm for datasource connection usage rate 

"DATASOURCE CONNECTION USAGE RATE" : Sends an alarm when the application's DataSource connection usage(%) exceeds the configured threshold."

PR #3116, opened 2017-07-06
[#3114] Add alarm for datasource connection usage rate

koo-taejin:
Add alarm for datasource connection usage rate


minwoo-jung:
good!

Issue #3211, opened 2017-08-02
add Netty plugin (support only 4.1.x.Final  version)


koo-taejin:
#### Description:
Add the netty plugin(for client). (support only 4.1.x.Final  version)
Netty trace is to record basic method.
needs each codec setting for support remote trace.

#### Result
* case : write in listener
```
Bootstrap bootstrap = getBootstrap();
bootstrap.connect(address).addListener(new ChannelFutureListener() {
    @Override
    public void operationComplete(ChannelFuture future) throws Exception {
        future.channel().writeAndFlush(request);
    }
});
```
![1](https://user-images.githubusercontent.com/10057874/28863188-f2990bba-77a2-11e7-9f3a-4e48b8b939a6.png)

* case: write 
```
Bootstrap bootstrap = getBootstrap();
bootstrap.connect(address).sync().channel();
channel.writeAndFlush(request);
```
![2](https://user-images.githubusercontent.com/10057874/28863311-5d9fc002-77a3-11e7-967f-cca3cb4a333d.png)

#### Additional Info
default setting is do not trace and record netty plugin
recommend  netty plugin disable, when using VERTX.

dinesh4747:
Hi @koo-taejin @Xylus 

In reference to the support of Netty framework does it also extends in tracing our Ratpack libraries (Since ratpack http method interface orginates from netty libraries)

koo-taejin:
Hi @dinesh4747 

Are you asking if pinpoint supports Ratpack?
If true, Pinpoint does not support Ratpack.

But, If Ratpack uses Netty 4.1.x version and use Netty's Channel and Netty's HttpCodec, then trace may be possible. (but i'm not sure.)

PR #3212, opened 2017-08-02
[#3211] add Netty plugin (support only 4.1.x.Final version)

koo-taejin:


plugins/netty/src/main/java/com/navercorp/pinpoint/plugin/netty/interceptor/http/HttpEncoderInterceptor.java
```diff
@@ -0,0 +1,318 @@
...
+
+        ChannelHandlerContext channelHandlerContext = (ChannelHandlerContext) args[0];
+        Channel channel = channelHandlerContext.channel();
+        final String endpoint = NettyUtils.getEndPoint(channel.remoteAddress());
+        recorder.recordDestinationId(endpoint);
+
+        HttpMessage httpMessage = (HttpMessage) args[1];
+        HttpHeaders headers = httpMessage.headers();
+
+        addHeaderIfNotExist(headers, Header.HTTP_SAMPLED.toString(), SamplingFlagUtils.SAMPLING_RATE_FALSE);
```
Xylus: Shouldn't this be set after checking for `Trace.canSampled()` is false?

koo-taejin: it's my mistake, i will fix it. thanks :)

PR #3213, opened 2017-08-03
Allows specifying multiple source/destination applications when querying for node histograms

Xylus:
Some nodes may have inbound edges from multiple applications. This allows the node histogram API to receive multiple applications for these cases.

### DIFF
--- a/web/src/main/java/com/navercorp/pinpoint/web/controller/MapController.java
+++ b/web/src/main/java/com/navercorp/pinpoint/web/controller/MapController.java
@@ -30,18 +30,24 @@ import com.navercorp.pinpoint.web.view.ApplicationTimeHistogramViewModel;
 import com.navercorp.pinpoint.web.applicationmap.nodes.NodeHistogramSummary;
 import com.navercorp.pinpoint.web.vo.Application;
 import com.navercorp.pinpoint.web.vo.ApplicationPair;
+import com.navercorp.pinpoint.web.vo.ApplicationPairs;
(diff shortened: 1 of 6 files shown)


web/src/main/java/com/navercorp/pinpoint/web/controller/MapController.java
```diff
@@ -256,54 +262,30 @@ public NodeHistogramSummary postResponseTimeHistogramDataV2(
...
-            toApplication = applicationFactory.createApplication(toApplicationName, applicationPair.getToServiceTypeCode());
-        }
+        List<Application> fromApplications = mapApplicationPairsToApplications(applicationPairs.getFromApplications());
+        List<Application> toApplications = mapApplicationPairsToApplications(applicationPairs.getToApplications());
 
-        return responseTimeHistogramService.selectNodeHistogramData(application, range, fromApplication, toApplication);
+        return responseTimeHistogramService.selectNodeHistogramData(application, range, fromApplications, toApplications);
     }
 
-    @RequestMapping(value = "/getResponseTimeHistogramDataV2", method = RequestMethod.GET)
```
koo-taejin: Will using the getResponseTimeHistogramDataV2 method with HTTP GET disappear?

Xylus: The plan was to use POST only as from / to application parameters may be too long for GETs.
But still having a GET API available would be a good idea,
Will add it back, thanks!

Issue #3275, opened 2017-08-24
Supports TCP transport type for Span and Stat data


koo-taejin:
#### Description
Currently, only UDP transport type is supported for Span and Stat data, 
Supports TCP transport type for Span and Stat data. (cause : some people wanted)

#### Result
* collector
Can determine the Span and Stat transport types in the collector.
The setting should be  TCP support, UDP support, or both of all support.
The settings are as follows.
```
collector.receiver.span.udp=true
collector.receiver.span.udp.ip=0.0.0.0
collector.receiver.span.udp.port=9996
collector.receiver.span.udp.receiveBufferSize=4194304

collector.receiver.span.tcp=false
collector.receiver.span.tcp.ip=0.0.0.0
collector.receiver.span.tcp.port=9996
```

* agent
Can determine the Span and Stat transport types in the agnet.
The setting should be  TCP support, UDP support.
The settings are as follows.
```
profiler.statdatasender.transport.type=UDP // or TCP
profiler.spandatasender.transport.type=UDP // or TCP
```

#### Additional Info
Should keep in mind that TCP transport load balancing is per connection. 
(UDP transport loadbalancing is per packet)

PR #3276, opened 2017-08-24
[#3275] Supports TCP transport type for Span and Stat data

koo-taejin:
1. implement collector side
2. add test case


minwoo-jung:
awesome, LGTM!

PR #3383, opened 2017-09-20
fix for jdk-8

umutcelik:


emeroad:
Sorry. We do not use unnecessary stream.

PR #3422, opened 2017-09-29
Udpdnscache

### LINKED ISSUE #3006: Agent Connectivity issue using ELB for collector
#### Description
 We ran into an issue with agents connecting via collector ELB  , found out that agents having a persistent connection to a ELB IP (collector ) . If the ELB IP changes for any reasons  the agent does not automatically connect to the new ELB IP unless we restart the agent . In the pinpoint.config, Alias name for the ELB used and NOT the IP address

#### Environment

* Pinpoint version:1.6.0
* Amazon Linux-4.9.27-14.31., 
* Additional details two web ec2 behind a ALB, 2 collector EC2 on behind a ELB:

#### Question 
We cannot restart the agent every time the ELB IP changes , is there a way to make the ELB IP's transparent to agents

teambullseye:
resolves #3006 
datagram socket connection is recreated every 5 mins


emeroad:
resolved by  #3827

PR #3501, opened 2017-10-31
fixed unknown Group Node names overlap in topology

bigsnapdragon:
bug repeat as follows:
1 first, status
1 first

2 clicked on the unknow group node
2clickon

3 and refresh the topology
3 refresh

4 then display overlap
4 overlap

To fix this problem, need to remove ‘category’ from the data json


denzelsN:
Group node UI is changed in master branch . 
So can not reproduce the bug.

Issue #3450, opened 2017-10-18
Remove duplicate data from agent stat chart APIs


Xylus:
#### Description
The current response format for agent stat chart APIs have a lot of duplicate data entries. It would be nice to get this remedied.
Before : 
```
{
  charts: {
    "TPS_SAMPLED_CONTINUATION": {
      "points": [{
        xVal: milisecond,
        minYVal: number,
        maxYVal: number,
        avgYVal: number,
        sumYVal: number  
      }, ...]
    },
    "TPS_SAMPLED_NEW": { ... },
    "TPS_TOTAL": { ... },
    "TPS_UNSAMPLED_CONTINUATION": { ... },
    "TPS_UNSAMPLED_NEW": { ... },
  }
}
```

After : 
```
{
  "schema": {
    "x": "time",
    "TPS_SAMPLED_CONTINUATION": ["min", "max", "avg", "sum"],
    "TPS_SAMPLED_NEW": ["min", "max", "avg", "sum"],
    "TPS_TOTAL": ["min", "max", "avg", "sum"],
    "TPS_UNSAMPLED_CONTINUATION": ["min", "max", "avg", "sum"],
    "TPS_UNSAMPLED_NEW": ["min", "max", "avg", "sum"], 
  },
  "x": [ time, time, time, ... ],
  "y": {
    "TPS_SAMPLED_CONTINUATION": [
       [ number, number, number, number ],
       ...
     ],
    "TPS_SAMPLED_NEW": [ ... ],
    "TPS_TOTAL": [ ... ],
    "TPS_UNSAMPLED_CONTINUATION": [ ... ],
    "TPS_UNSAMPLED_NEW": [ ... ],
  }
}
```

PR #3502, opened 2017-10-31
Removes duplicate data from stat chart APIs

Xylus:
This cleans up lots of duplicate data (such as timestamp, keys, etc) from both agent and application stat chart APIs.
Consequently, to homogenize the responses, agent and application stat charts have been refactored into a common interface.

related to #3450


web/src/main/java/com/navercorp/pinpoint/web/view/AgentStatPointSerializer.java
```diff
@@ -0,0 +1,56 @@
...
+    @Deprecated
+    @Value("#{pinpointWebProps['web.stat.chart.version'] ?: 'v1'}")
+    private String version;
+
+    @Override
+    public void serialize(AgentStatPoint<? extends Number> agentStatPoint, JsonGenerator jgen, SerializerProvider serializers) throws IOException, JsonProcessingException {
+        if ("v1".equalsIgnoreCase(version)) {
+            jgen.writeStartObject();
+            jgen.writeNumberField("xVal", agentStatPoint.getxVal());
+            jgen.writeObjectField("minYVal", agentStatPoint.getMinYVal());
```
koo-taejin: I think It is better to use writeNumber method than writeObject method. :)

Issue #3580, opened 2017-11-21
1.7.0-RC1 calls api.github.com and fails when unreachable


ofauchon:
2017-11-21 09:37:14 [INFO ](AgentStatDaoFactory$JvmGcDaoFactory) AgentStatDao Compatibility v2
2017-11-21 09:37:15 [WARN ](utowiredAnnotationBeanPostProcessor) Autowired annotation should only be used on methods with parameters: public java.util.List com.navercorp.pinpoint.web.dao.rest.GithubAgentDownloadInfoDao.getDownloadInfoList()
2017-11-21 09:37:15 [WARN ](o.s.w.c.s.XmlWebApplicationContext ) Exception encountered during context initialization - cancelling refresh attempt: org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'agentInfoServiceImpl': Unsatisfied dependency expressed through field 'agentDownloadInfoDao'; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'agentDownloadInfoDao': Injection of autowired dependencies failed; nested exception is org.springframework.web.client.ResourceAccessException: I/O error on GET request for "https://api.github.com/repos/naver/pinpoint/releases": api.github.com; nested exception is java.net.UnknownHostException: api.github.com
2017-11-21 09:37:15 [INFO ](onManager$HConnectionImplementation) Closing zookeeper sessionid=0x15fca95f3a20017
2017-11-21 09:37:15 [INFO ](o.a.h.h.s.o.a.z.ZooKeeper          ) Session: 0x15fca95f3a20017 closed
2017-11-21 09:37:15 [INFO ](o.a.h.h.s.o.a.z.ClientCnxn         ) EventThread shut down
2017-11-21 09:37:15 [INFO ](c.n.p.c.h.HbaseTemplate2           ) HBaseTemplate2.destroy()
2017-11-21 09:37:15 [WARN ](c.n.p.c.h.HbaseTemplate2           ) count -1
2017-11-21 09:37:15 [INFO ](c.n.p.c.h.PooledHTableFactory      ) PooledHTableFactory.destroy()
2017-11-21 09:37:15 [INFO ](onManager$HConnectionImplementation) Closing zookeeper sessionid=0x15fca95f3a20016
2017-11-21 09:37:15 [INFO ](o.a.h.h.s.o.a.z.ZooKeeper          ) Session: 0x15fca95f3a20016 closed
[...]

PR #3586, opened 2017-11-23
[#3580] Fixed accessing github api at startup in web

koo-taejin:
1. add download info configuration
2. Makes api call lazilly


koo-taejin:
@emeroad review please :)

Issue #3613, opened 2017-12-06
It's better to use catalina.base instead of catalina.home


hanyong:
#### Description

I have tomcat 7 installed and configured for multiple instances, CATALINA_HOME is readonly and shared by several apps, each app have it's own CATALINA_BASE. When I try to run collector in one CATALINA_BASE, the path for `stat.log` is set in CATALINA_HOME, and is not suitable and not writable.

We can see the error message from `catalina.out`:

```
Dec 06, 2017 4:35:29 PM org.apache.catalina.startup.HostConfig deployWAR
INFO: Deploying web application archive /home/observer.hany/opt/pinpoint-collector/webapps/ROOT.war
log4j:ERROR setFile(null,true) call failed.
java.io.FileNotFoundException: /opt/apache-tomcat-7.0.56/logs/stat.log (权限不够)
	at java.io.FileOutputStream.open0(Native Method)
	at java.io.FileOutputStream.open(FileOutputStream.java:270)
```

It's better to use `catalina.base` instead of `catalina.home` for logs.

#### Environment

* Pinpoint version: 1.7.0-SNAPSHOT
* affected library, etc: collector
* Additional details (e.g. how your collector/web is set up):

minwoo-jung:
Hi hanyong
Please see [log4j.xml configuration file](https://github.com/naver/pinpoint/blob/master/collector/src/main/resources-release/log4j.xml)

If you want to set directory for stat file, you edit log4j.xml configuration.

minwoo-jung:
@hanyong 
Instead of You sending PR, you should just edit your local code.

hanyong:
Hi, @minwoo-jung 

We can run tomcat in two mode (See https://tomcat.apache.org/tomcat-7.0-doc/appdev/installation.html):

1. Single instance, the CATALINA_HOME is also the CATALINA_BASE.
2. Multiple instances，each instance will have its own CATALINA_BASE configured.

Use `catalina.home` only works for the first mode, use `catalina.base` works for both mode, and also works for tomcat-maven-plugin. So `catalina.base` have better compatibility. That's why I sending the PR.

hanyong:
For more portable, maybe we should set and use something like `pinpoint.home` ?

PR #3614, opened 2017-12-06
[#3613] use "catalina.base" for logs path

hanyong:
use "catalina.base" instead of "catalina.home" for logs path


minwoo-jung:
@hanyong

PR #3662, opened 2017-12-28
add .DS_Store for macOS

mysqto:
For those who using macOS, .DS_Store is always annoying


emeroad:
Fixed in https://github.com/naver/pinpoint/commit/f28c16536b273b2ee5563084ea05f0ebef7bc7fa

Issue #3647, opened 2017-12-21
Question for deadLock


Super4You:
![deadlock](https://user-images.githubusercontent.com/25604732/34261455-216bdfec-e6a4-11e7-8bca-c3b6dd71d556.png)
Hello，I'm sorry to bother you，When I added a agent for an application,  at startup, the Main thread will initial redis cache so that transform in the pinpoint-redis-plugin will be called.This time the class of Jedis was loaded by AppClassloader,but  in the process of execute transform pinpoint-redis-plugin may use PinpointURLClassLoader load some class, at the same time, The thread Pinpoint-Client-Worker first to get the lock of loadClass method , further access to lock of  Classloader method  loadClass, There is a deadlock in threads of Main and Pinpoint-Client-Worker. Attached screenshot for jstack  information, may have deviation analysis, so please give a guide,if this issue is a bug.

Then,there is anthor question
![11111](https://user-images.githubusercontent.com/25604732/33629765-34054492-da40-11e7-87c2-442538737cd3.png)
![22222](https://user-images.githubusercontent.com/25604732/33629767-344dfc50-da40-11e7-834c-cf0e879b166a.png)
As shown, because of the framework are all asynchronous implementation, the effect of the display is the time consuming of each asynchronous module, Whether this time  statistics  one request total time?If there is any problem with my plugin, Look forward to your letter.Thank you.

koo-taejin:
hi Super4You

### Q1
Currently, there is a potential risk that a pinpoint will deadlock if you use jvm version less 6 or lower.
However, in case of 7 or more, there is no problem using parallel classloader.g jdk7 or later.
However, I am working on jdk6 to avoid deadlock as much as possible, and I will work on improvement in this issue.
I will ask you for a verification request after the work is over.
 
### Q2
exec time, is the time from the beginning to the end of the method.

What is the request total time you said?
1. The time is request ends in the client
// look at execTime of sendAsync
2. The time when the thread that first received the request from the server ends
// look at  execTime of afterRececeived method
3. The time at which all tasks executed by the request on the server end
// Look at the time of the last asynchronous invocation

I don't know the structure of the plugin you created and can not give the correct answer, but it looks great.

thanks.

koo-taejin:
hi @Super4You 

added function of selecting profiler's parent's classloader.
Through this function, I expect that deadlock will not occur.

TODO : must be set following option
```
# select parent classloader (BOOTSTRAP / EXTENSION / SYSTEM)
[...]

PR #3676, opened 2018-01-04
[#3647] prevent deadlock when load class in jdk6

koo-taejin:
1. support to select profiler parent's classloader


emeroad:
fixed #4051

PR #3723, opened 2018-01-17
Adds hbase namespace support

Xylus:
Allows collector, web, and flink modules to configure hbase namespace.
Default value is blank (which defaults to *default* namespace) which is what pinpoint is currently using.


DiligenceY:
is it ok? waiting for this

Xylus:
@DiligenceY Just doing some last minute testing. Should be merged early next week if everything checks out

Issue #3729, opened 2018-01-18
Refactor the collector threadpool


emeroad:
### Description
 - Remove duplicate code
 - Simplifies threadpool handling
 - Change to spring-based configuration

PR #3731, opened 2018-01-18
[#3729] Refactor the collector threadpool

emeroad:
#3729


collector/src/main/java/com/navercorp/pinpoint/collector/receiver/udp/UDPReceiver.java
```diff
@@ -221,14 +216,9 @@ private void bindSocket(DatagramSocket socket, InetSocketAddress bindAddress) {
     }
 
     private int getPacketPoolSize() {
-        int threadSize = worker.getThreadSize();
-        int queueSize = worker.getQueueSize();
-
-        return threadSize + queueSize + ioThreadSize;
+        return 1024*5;
```
koo-taejin, line 219: Is it fixed value?

emeroad: This is a temporary code
In the next issue, I will extract the datagram pool into a Spring-Bean

koo-taejin: good :)

Issue #3747, opened 2018-01-23
Allows the profiler to change the guice's module creation


koo-taejin:
#### Description
Allows the profiler to change the guice's module creation.
It can create own modules and change the code for the main logic of the profiler.

#### Additional Info(configuration)
 ```
// in pinpoint.config (default : empty)
profiler.guice.module.factory=className
```

PR #3748, opened 2018-01-23
[#3747] Allows the profiler to change the guice's module creation

koo-taejin:
Allows the profiler to change the guice's module creation


koo-taejin:
@emeroad review please :)

Issue #3177, opened 2017-07-21
Plan to impletation of Per-request logging


dayforward:
we are going to finish the Per-request logging based on tag of 1.6.2  of pinpoint project in about half a month.In our implementation, we target at collecting business log at the dir of log of tomcat.
Util now we've already implements most coding including collecting logs 
in agent, putting logs into hbase , and displaying in the webui. 
our new branch is located at https://github.com/dayforward/pinpoint
Welcome anyone inttresting in this to give advice. 
#### Environment

* Pinpoint version:1.6.2

minwoo-jung:
looks good, good luck!

dayforward:
[businesslog.zip](https://github.com/naver/pinpoint/files/1346109/businesslog.zip)
I'm sorry for this delayed patch for businesslog implementation; 
The patch is based on tag of 1.6.2 of pinpoint.  I'll be  very pleased if our community 
to accept the patch ,yet it needs lots of impovements in future. if not , I 'll keep on 
improve the patch ....

dayforward:
@minwoo-jung, I've impleted an initial version of pre-request and already keeping up with the master branch. Would you please do an code review , thanks for your reply!

minwoo-jung:
Hi @dayforward 
I got it. I will review and give you feedback.

minwoo-jung:
Hi @dayforward

I reviewed your code at your repository. 
I should highly praise your efforts. I think you understand pinpoint core logic pretty well.

I'm tried to execute your per-request function. I wish you explain how to execute per-request implement function.
such as profiler.tomcatlog.dir property and log4j.xml configuration, and ConversionPattern.
**I think it would be best to add on sample project to your repository branch.**

And i want to see image in readme file , But image is broken link.

dayforward:
Hi，Minwoo Jung
Sorry for my late reply for We have our Chinese New Year these days.
I've pushed an example in the github https://github.com/dayforward/pinpoint/tree/log4jexample_for_prerequest 
this is an tomcat project , and README is also pushed in the log4jexample_for_prerequest branch, thanks for your review.



sunyangzju@163.com
 
From: Minwoo Jung
Date: 2018-02-21 13:34
To: naver/pinpoint
CC: dayforward; Mention
Subject: Re: [naver/pinpoint] Plan to impletation of Per-request logging (#3177)
Hi @dayforward
I reviewed your code at your repository.
I should highly praise your efforts. I think you understand pinpoint core logic pretty well.
[...]

PR #3792, opened 2018-02-02
impletation of pre-request

dayforward:
impletation of pre-request


dayforward:
related issue is "Plan to impletation of Per-request logging #3177"

Issue #3733, opened 2018-01-18
TCP reconnect issue when init with domain


gaoyf:
I'm sorry to bother you again, but I encounter another question.
My agent config like bellow:
`profiler.collector.ip=trace.xxx.com`

And my logic deployment like bellow:
![](https://user-images.githubusercontent.com/10137071/34399808-06a525e8-ebc6-11e7-96af-21ce24222211.png)

And my question just like ![#3658](https://github.com/naver/pinpoint/issues/3658), but its about TCP, as bellow: 
**When the ip list of domain name change, the tcp socket still connects to the old ip .**
Bellow is the agent's log:
```
2018-01-18 10:28:24 [WARN ](tPinpointClientFactory$ConnectEvent) try reconnect. connectAddress:trace.tv.sohuno.com/10.10.11.16:9994
2018-01-18 10:28:32 [WARN ](tPinpointClientFactory$ConnectEvent) try reconnect. connectAddress:trace.tv.sohuno.com/10.10.11.16:9994
2018-01-18 10:28:40 [WARN ](tPinpointClientFactory$ConnectEvent) try reconnect. connectAddress:trace.tv.sohuno.com/10.10.11.16:9994
```

I checked `ClientFactoryUtils` code, found that:
```
public static PinpointClient createPinpointClient(String host, int port, PinpointClientFactory clientFactory) {
    InetSocketAddress connectAddress = new InetSocketAddress(host, port);
    return createPinpointClient(connectAddress, clientFactory);
}
```
**The key point is the hostname only be determined once when new InetSocketAddress  Object!**

This question is like ![ZOOKEEPER-1846: Always recreate socket addresses for any DNS change](https://github.com/apache/zookeeper/pull/37).

And I known, it's not recommend to use L7 as load balancer, but I'm familiar with nginx only. 
So do you have a plan to improve the compatibility when use L7 LB?  Or fix the isssue like zookeeper?
Thank you very much!

emeroad:
I fully agree with your opinion.
We will fix the problem quickly.

Thank you for your feedback.

gaoyf:
:thumbsup: for your quickly response and responsible attitude!

PR #3801, opened 2018-02-05
[#3733] Fix DNS resolving for tcp

emeroad:
- Refactor rpc client


rpc/src/main/java/com/navercorp/pinpoint/rpc/client/Connection.java
```diff
@@ -0,0 +1,135 @@
...
+        handlerConnectFuture.awaitUninterruptibly();
+
+        if (ConnectFuture.Result.FAIL == handlerConnectFuture.getResult()) {
+            throw new PinpointSocketException("connect fail to " + getRemoteAddress() + ".", connectFuture.getCause());
+        }
+
+        return pinpointClientHandler;
+    }
+
+    public PinpointClient awaitHandshake() {
```
koo-taejin: i think method name(awaitHandshakeStarted) seems to be better

commons/src/main/java/com/navercorp/pinpoint/common/util/NetUtils.java
```diff
@@ -36,32 +36,77 @@
...
+    }
 
-            return new InetSocketAddress(uri.getHost(), uri.getPort());
-        } catch (URISyntaxException ignore) {
-            // skip
+    public  static <T> T parseHostAndPort(String address, HostAndPortFactory<T> hostAndPortFactory) {
+        if (StringUtils.isEmpty(address)) {
+            return null;
         }
+        Assert.requireNonNull(address, "address must not be null");
```
koo-taejin: It looks like duplicate validation.
If it is not a duplicate validation, then order should change.
or It might be a validation for hostAndPortFactory variable.

emeroad: opps! 👍

Issue #3664, opened 2017-12-29
Add hbase namespace support


Xylus:
#### Description
Adding namespaces would add a layer of logical grouping of tables that could be useful for multi-tenancy related features going forward for cases described in #3535

PR #3808, opened 2018-02-06
Separates namespace wiring configuration

Xylus:
related to #3664


Xylus:
@minwoo-jung review please :)

PR #3812, opened 2018-02-07
Applies context root for quickstart modules

Xylus:
* Health check urls in startup scripts takes context path into account if configured.
* Links in test app UI are generated properly with context path prepended.


Xylus:
@koo-taejin review please :)

PR #3814, opened 2018-02-08
impletation of Per-request logging

dayforward:
related issue is "Plan to impletation of Per-request logging #3177"


minwoo-jung:
https://github.com/naver/pinpoint/issues/3177#issuecomment-373271491

Issue #3926, opened 2018-03-21
[rpc] Add authentication handler interface


koo-taejin:
#### Description
Authentication may be needed at first connection. 
To support this, create authentication handler interface.

This issue is create interface for when authentication is required.
It should be implemented by profiler, collector, and web according to the authentication type.

#### Additional Info
AuthenticationManager interface
```
public interface AuthenticationManager {

    AuthenticationResult authenticate(Authentication authentication, ChannelHandlerContext channelHandlerContext);

}

```

PR #3927, opened 2018-03-21
[#3926] Add authentication handler interface

koo-taejin:
Add authentication handler interface


koo-taejin:
@emeroad review please :)

Issue #3930, opened 2018-03-22
add to websocketHandlerDecoratorFactory 


minwoo-jung:
if you want to wap it When add WebSocketHandler to WebSocketHandlerRegistry ,
You be able to generate and add websocketHandlerDecorator class.

PR #3931, opened 2018-03-22
[#3930] add to websocketHandlerDecoratorFactory for wapping webSocket…

minwoo-jung:


web/src/main/java/com/navercorp/pinpoint/web/websocket/WebSocketHandlerDecoratorFactory.java
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+package com.navercorp.pinpoint.web.websocket;
+
+import org.springframework.web.socket.WebSocketHandler;
+
+/**
+ * @author minwoo.jung
+ */
+public interface WebSocketHandlerDecoratorFactory {
```
Xylus, line 23: There's a **org.springframework.web.socket.handler.WebSocketHandlerDecoratorFactory** interface that we can use instead of creating a new one?

minwoo-jung: Ok~ I agree with you.

Issue #3906, opened 2018-03-15
Dubbo application, PinpointException: already Trace Object exist


aiqing2171:
**version:pinpoint 1.7.2**
2018-03-15 17:47:20 [WARN ](.n.p.p.d.i.DubboProviderInterceptor) [TxId:,SpanId:] BEFORE. Caused**:already Trace Object exist**.
com.navercorp.pinpoint.exception.PinpointException: already Trace Object exist.
        at com.navercorp.pinpoint.profiler.context.DefaultTraceFactory.checkAndGet(DefaultTraceFactory.java:102)
        at com.navercorp.pinpoint.profiler.context.DefaultTraceFactory.newTraceObject(DefaultTraceFactory.java:113)
        at com.navercorp.pinpoint.profiler.context.DefaultTraceContext.newTraceObject(DefaultTraceContext.java:137)
        at com.navercorp.pinpoint.plugin.dubbo.interceptor.DubboProviderInterceptor.createTrace(DubboProviderInterceptor.java:43)
        at com.navercorp.pinpoint.bootstrap.interceptor.SpanSimpleAroundInterceptor.before(SpanSimpleAroundInterceptor.java:51)
        at com.navercorp.pinpoint.bootstrap.interceptor.ExceptionHandleAroundInterceptor.before(ExceptionHandleAroundInterceptor.java:35)
        at com.alibaba.dubbo.rpc.proxy.AbstractProxyInvoker.invoke(AbstractProxyInvoker.java)
        at com.alibaba.dubbo.rpc.protocol.InvokerWrapper.invoke(InvokerWrapper.java:53)
        at com.alibaba.dubbo.rpc.filter.ExceptionFilter.invoke(ExceptionFilter.java:64)
        at com.alibaba.dubbo.rpc.protocol.ProtocolFilterWrapper$1.invoke(ProtocolFilterWrapper.java:91)
        at com.alibaba.dubbo.rpc.filter.TimeoutFilter.invoke(TimeoutFilter.java:42)
        at com.alibaba.dubbo.rpc.protocol.ProtocolFilterWrapper$1.invoke(ProtocolFilterWrapper.java:91)
        at com.*.wpp.tracing.dubbo.filter.TracingFilter.invoke(TracingFilter.java:123)  _---it implements dubbo.rpc.Filter,used for logging dubbo interface info_
[...]

Issue #3318, opened 2017-09-07
hello,dubbo client pinpoint agent Throw exception: already Trace Object exist. What is the reason?


hanrongsu:
hi!
agent version: 1.6.0
dubbo version: dubbo-2.8.3.jar

2017-09-07 10:31:10 [WARN ](.n.p.p.d.i.DubboProviderInterceptor:406) BEFORE. Caused:already Trace Object exist.
com.navercorp.pinpoint.exception.PinpointException: already Trace Object exist.
	at com.navercorp.pinpoint.profiler.context.ThreadLocalTraceFactory.checkBeforeTraceObject(ThreadLocalTraceFactory.java:146)
	at com.navercorp.pinpoint.profiler.context.ThreadLocalTraceFactory.continueTraceObject(ThreadLocalTraceFactory.java:120)
	at com.navercorp.pinpoint.profiler.context.active.ActiveTraceFactory.continueTraceObject(ActiveTraceFactory.java:86)
	at com.navercorp.pinpoint.profiler.context.DefaultTraceContext.continueTraceObject(DefaultTraceContext.java:159)
	at com.navercorp.pinpoint.plugin.dubbo.interceptor.DubboProviderInterceptor.createTrace(DubboProviderInterceptor.java:50)
	at com.navercorp.pinpoint.bootstrap.interceptor.SpanSimpleAroundInterceptor.before(SpanSimpleAroundInterceptor.java:51)
	at com.navercorp.pinpoint.bootstrap.interceptor.ExceptionHandleAroundInterceptor.before(ExceptionHandleAroundInterceptor.java:35)
	at com.alibaba.dubbo.rpc.proxy.AbstractProxyInvoker.invoke(AbstractProxyInvoker.java)
	at com.alibaba.dubbo.rpc.protocol.AbstractProxyProtocol$2.doInvoke(AbstractProxyProtocol.java:93)
	at com.alibaba.dubbo.rpc.protocol.AbstractInvoker.invoke(AbstractInvoker.java:144)
	at com.alibaba.dubbo.rpc.listener.ListenerInvokerWrapper.invoke(ListenerInvokerWrapper.java:74)
	at com.alibaba.dubbo.rpc.protocol.dubbo.filter.FutureFilter.invoke(FutureFilter.java:53)
	at com.alibaba.dubbo.rpc.protocol.ProtocolFilterWrapper$1.invoke(ProtocolFilterWrapper.java:91)
[...]

Issue #2981, opened 2017-06-01
already Trace Object exist


sungy2014:
2017-05-25 18:39:02 [WARN ](.n.p.p.d.i.DubboProviderInterceptor) BEFORE. Caused:already Trace Object exist.
com.navercorp.pinpoint.exception.PinpointException: already Trace Object exist.
	at com.navercorp.pinpoint.profiler.context.ThreadLocalTraceFactory.checkBeforeTraceObject(ThreadLocalTraceFactory.java:142)
	at com.navercorp.pinpoint.profiler.context.ThreadLocalTraceFactory.newTraceObject(ThreadLocalTraceFactory.java:152)
	at com.navercorp.pinpoint.profiler.context.active.ActiveTraceFactory.newTraceObject(ActiveTraceFactory.java:100)
	at com.navercorp.pinpoint.profiler.context.DefaultTraceContext.newTraceObject(DefaultTraceContext.java:174)
	at com.navercorp.pinpoint.plugin.dubbo.interceptor.DubboProviderInterceptor.createTrace(DubboProviderInterceptor.java:41)
	at com.navercorp.pinpoint.bootstrap.interceptor.SpanSimpleAroundInterceptor.before(SpanSimpleAroundInterceptor.java:51)
	at com.navercorp.pinpoint.bootstrap.interceptor.ExceptionHandleAroundInterceptor.before(ExceptionHandleAroundInterceptor.java:35)
	at com.alibaba.dubbo.rpc.proxy.AbstractProxyInvoker.invoke(AbstractProxyInvoker.java)
	at com.alibaba.dubbo.rpc.protocol.AbstractProxyProtocol$2.doInvoke(AbstractProxyProtocol.java:93)
	at com.alibaba.dubbo.rpc.protocol.AbstractInvoker.invoke(AbstractInvoker.java:144)
	at com.alibaba.dubbo.rpc.listener.ListenerInvokerWrapper.invoke(ListenerInvokerWrapper.java:74)
	at com.mockuai.common.dubbo.filter.TraceIdConsumerFilter.invoke(TraceIdConsumerFilter.java:43)
	at com.alibaba.dubbo.rpc.protocol.ProtocolFilterWrapper$1.invoke(ProtocolFilterWrapper.java:91)
	at com.alibaba.dubbo.monitor.support.MonitorFilter.invoke(MonitorFilter.java:75)
[...]

PR #3971, opened 2018-04-03
[#3906] dubbo plugin: already trace exist

jiaqifeng:
When dubbo provider try to create trace, there maybe already has a trace in threadlocal. To avoid re-create a new trace, add a scope to trace to indicate not to remove the trace when leaving dubbo provider method.
This should fix issue #3906, #2981 and #3318.
To reproduce the condition I only using injvm call mode of dubbo to verify the fix, but I am not sure this will cover all 3 issues mentions above. The test app is at https://github.com/jiaqifeng/quickdemo.


jaehong-kim:
@jiaqifeng 
Thanks for the contribution.
There is no logic to delete variables entered in the scope.
It is dangerous if it is a recursive call.
Please see this source. 
https://github.com/naver/pinpoint/blob/master/plugins/user/src/main/java/com/navercorp/pinpoint/plugin/user/interceptor/UserIncludeMethodInterceptor.java

Xylus:
@jiaqifeng 
Thanks for the fix!
Will merge first and fix any [potential issue](https://github.com/naver/pinpoint/pull/3971#issuecomment-381514600) afterwards.

Issue #3995, opened 2018-04-09
rabbitmq plugin did not passed traceId for arg validation failure


jiaqifeng:
I tested lastest Rabbitmq Plugin on my app from https://github.com/jiaqifeng/quickdemo, and found the traceId did passed from sender to receiver. The root casuse is ChannelBasicPublishInterceptor.validate() will return false when args[n] is null whatever it's class is.
I fixed it and PR will be submitted.

jiaqifeng:
PR is 3996

Xylus:
@jiaqifeng Thanks!
To clarify, this is an issue where trace id is currently not passed (and publisher not traced) due to validation failure when *exchange*, or *props*, etc are null, right? And that it should still trace and pass trace id even if some of the parameters are null?

jiaqifeng:
@Xylus This bug will cause sender did not put the txId into rabbitmq message and receiver will create a new txId. So in topology map the sender and receiver will not be linked together, sender is linked to virtual queue node only. So does receiver node.

PR #3996, opened 2018-04-09
[#3995] rabbitmq plugin not pass traceId

jiaqifeng:
If args[n]==null then instanceof will always fail, so check it before instanceof.

### DIFF
--- a/plugins/rabbitmq/src/main/java/com/navercorp/pinpoint/plugin/rabbitmq/client/interceptor/ChannelBasicPublishInterceptor.java
+++ b/plugins/rabbitmq/src/main/java/com/navercorp/pinpoint/plugin/rabbitmq/client/interceptor/ChannelBasicPublishInterceptor.java
@@ -171,9 +171,6 @@ public class ChannelBasicPublishInterceptor implements AroundInterceptor {
         if (args[4] !=null && !(args[4] instanceof AMQP.BasicProperties)) {
             return false;
         }
-        if (args[5] !=null && !(args[5] instanceof byte[])) {
-            return false;
-        }
         return true;
     }
 }
--- a/plugins/rabbitmq/src/main/java/com/navercorp/pinpoint/plugin/rabbitmq/client/interceptor/ChannelBasicPublishInterceptor.java
+++ b/plugins/rabbitmq/src/main/java/com/navercorp/pinpoint/plugin/rabbitmq/client/interceptor/ChannelBasicPublishInterceptor.java
@@ -162,16 +162,16 @@ public class ChannelBasicPublishInterceptor implements AroundInterceptor {
         if (args == null || args.length < 6) {
             return false;
         }
-        if (!(args[0] instanceof String)) {
+        if (args[0] !=null && !(args[0] instanceof String)) {
             return false;
         }
-        if (!(args[1] instanceof String)) {
+        if (args[1] !=null && !(args[1] instanceof String)) {
             return false;
         }
-        if (!(args[4] instanceof AMQP.BasicProperties)) {
+        if (args[4] !=null && !(args[4] instanceof AMQP.BasicProperties)) {
             return false;
         }
-        if (!(args[5] instanceof byte[])) {
+        if (args[5] !=null && !(args[5] instanceof byte[])) {
             return false;
         }
         return true;


plugins/rabbitmq/src/main/java/com/navercorp/pinpoint/plugin/rabbitmq/client/interceptor/ChannelBasicPublishInterceptor.java
```diff
@@ -162,16 +162,16 @@ private boolean validate(Object target, Object[] args) {
         if (args == null || args.length < 6) {
             return false;
         }
-        if (!(args[0] instanceof String)) {
+        if (args[0] !=null && !(args[0] instanceof String)) {
             return false;
         }
-        if (!(args[1] instanceof String)) {
+        if (args[1] !=null && !(args[1] instanceof String)) {
```
Xylus, line 168: Allowing **null** routingKey will open up a window for NPE in the `after()` method above :
```
String routingKey = (String) args[1];
...
    ...
    recorder.recordAttribute(RabbitMQClientConstants.RABBITMQ_ROUTINGKEY_ANNOTATION_KEY, routingKey);
```

I think you should decide whether to skip or trace when routingKey is **null**.
If you'd like to skip, then I think validation should fail when `args[1] == null`.
If you decide to trace, then routingKey should be converted to a non-null (an empty string?) above.

edit: `recorder.recordAttribute(RabbitMQClientConstants.RABBITMQ_ROUTINGKEY_ANNOTATION_KEY, routingKey);` will throw NPE if `routingKey` is null.
```
public void recordAttribute(AnnotationKey key, String value) {
    addAnnotation(new Annotation(key.getCode(), value));
}
```
```
public Annotation(int key, String value) {
    super(key);
    this.setValue(TAnnotationValue.stringValue(value));
}
```
```
public static TAnnotationValue stringValue(java.lang.String value) {
      TAnnotationValue x = new TAnnotationValue();
      x.setStringValue(value);
      return x;
}
```
```
public void setStringValue(java.lang.String value) {
    if (value == null) throw new java.lang.NullPointerException();
    setField_ = _Fields.STRING_VALUE;
    value_ = value;
}
```

plugins/rabbitmq/src/main/java/com/navercorp/pinpoint/plugin/rabbitmq/client/interceptor/ChannelBasicPublishInterceptor.java
```diff
@@ -162,16 +162,16 @@ private boolean validate(Object target, Object[] args) {
...
             return false;
         }
-        if (!(args[4] instanceof AMQP.BasicProperties)) {
+        if (args[4] !=null && !(args[4] instanceof AMQP.BasicProperties)) {
             return false;
         }
-        if (!(args[5] instanceof byte[])) {
+        if (args[5] !=null && !(args[5] instanceof byte[])) {
```
Xylus: I think this whole check for `args[5]` can be removed since it's not used at all in the interceptor.
What do you think?

jiaqifeng: @Xylus In my test only args[4] instanceof AMQP.BasicProperties will be null. Since I did not know whether other args maybe null then I add check for all args. I know your concerns, for any args maybe null, the validation failure is easy to be ignored by user. If in some special case args[5] or [1] is null, exception will be printed into log which may cause we found the bug more easier and quicker. So it depends what you prefer.
At least, arg[4] should be check for null.

Xylus: @jiaqifeng I definitely agree with you that *args[4]* should be allowed to be null.

But regarding *args[1]* and *args[5]*, I think it's important to remember that validation check in interceptor is done to see if the parameters won't throw exceptions in the interceptor code **only**, and not the original method.
For example, if *args[1]* (routing key) is null, **AMQImpl$Basic$Publish** constructor will throw an **IllegalStateException** if it is null, but I don't think it's the interceptor's responsibility to tell the user that *args[1]* should not be null. (This is more to do with the comment above regarding *args[1]*)
Similarly, I don't think we need to do any checks for *args[5]* at all in the interceptor as we never access it to do anything in the interceptor code, but do let me know if you think the check is still useful.

jiaqifeng: @Xylus I agree with you. For I am not working on rabbitmq plugin for several months and forgot most details of rabbitmq client package's code.

Xylus: @jiaqifeng I think this part of your commit is just fine and can be merged as is :+1: 
Could you take a look at the other [comment](https://github.com/naver/pinpoint/pull/3996#discussion_r179973520)? That one does require a change I think.

plugins/rabbitmq/src/main/java/com/navercorp/pinpoint/plugin/rabbitmq/client/interceptor/ChannelBasicPublishInterceptor.java
jiaqifeng: args[0] is exhange and args[1] is routingKey, these two args should not be null which will cause exception by amqp-client 3.5.5. So it would not cause NPE in after(). So we need not check them against null.

plugins/rabbitmq/src/main/java/com/navercorp/pinpoint/plugin/rabbitmq/client/interceptor/ChannelBasicPublishInterceptor.java
jiaqifeng: @Xylus 
As other comment, args[0] and args[1] are exchange and routingKey which will be check not to be null by ampq-client 3.5.5, then we do not need to check them against null. For args[4], which will breake the tx chain if null, so it shoudl be check against null. For args[5] as body which is alowed to be null in amqp-client 3.5.5 code, and it is not used in pinpoint code so I suggest to ignore it. Do you agree this solution? If yes, I will update the branch.

plugins/rabbitmq/src/main/java/com/navercorp/pinpoint/plugin/rabbitmq/client/interceptor/ChannelBasicPublishInterceptor.java
Xylus: @jiaqifeng Ah yes, I see what you're saying.
NPE will indeed be thrown before after(), so yes you're right!
sorry for the confusion.

plugins/rabbitmq/src/main/java/com/navercorp/pinpoint/plugin/rabbitmq/client/interceptor/ChannelBasicPublishInterceptor.java
Xylus: @jiaqifeng yeh, I agree.
args[0], args[1], args[4] checks all look good.
The check for args[5] (body) on the other hand could be removed as it doesn't have anything to do with the interceptor.
Let me know if you decide to remove the check for args[5], if not I think this PR can be merged as is.

jiaqifeng: @Xylus 
I decided to remove check for args[5] and updated the branch already.

Xylus:
Thanks for the fix :)

PR #4136, opened 2018-05-17
Add Thrift installation script for Linux

chaerin16:
Hello.
I want to suggest an enhancement to resolve inconvenience when using Pinpoint.

I wanted to add monitoring metric, so I modified *.thrift file and builded it using build-thrift-linux.sh.
When I ran the build script, I realized there is no Thrift executable binary for Linux in Pinpoint.
So, I tried installing Thrift.
Uncomfortably, I had to get a specific Thrift version according to Pinpoint version. 
And I had to find how to install Thrift according to my Linux distribution and version.

Linux users will try running the build script simply, but they will be met error because there is no executable binary.
So, they have to find a specific version of Thrift, install it, and copy it to specific location in Pinpoint project.
It's very annoying.

So, I suggest an enhancement.
For Linux, locate a Thrift tarball whose version corresponds a specific version of Pinpoint in 'pinpoint/thrift/src/compiler/linux' path.
And build-thrift-linux.sh checks if Thrift is installed or not.
If it's not installed, it calls Thrift installation script.

This PR contains a installation script 'install-thrift-linux.sh' for Redhat/CentOS 7 as a prototype. (Supported Linux destribution can be added.)
And I modified build-thrift-linux.sh to check installation of Thrift and to call the installation script if it is necessary.

If this enhancement is accepted, many Pinpoint developer using Linux will be comfortable to add monitoring metric.

Thank you.


koo-taejin:
hi chaerin16

Really appreciate your contribution and suggestion.

I will check you're pull request.

thanks.

koo-taejin:
@chaerin16 

I've tested it,
I learned that many exceptions can occur.
 
I hope I can install it even if I am not root user. (If the required library is already installed)
But I know this is difficult. :(

Could you add  below things?
  1. Do not run install-thrift-linux.sh if it already exists.
  2. Doing running install-thrift-linux.sh
  ```
  installed_path = `which thrift` 
  ```
 Checks if install completed or failed.
 and if failed, then show error message with thrift compiler install page link.
 
thanks.

chaerin16:
@koo-taejin

Thank you for your checking.
I will add code you mentioned and upload a patch.

Thank you.

chaerin16:
Hello.

I modified codes as following and committed it.

1. do not install thrift when it installed in system already.
2. do not require root authority when pre-installed packages is installed already.
3. show thrift install page link when installation failed.

In the beginning, I thought the item 2 is not possible because 'make install' require root authorify.
But I discovered there is no problem using thrift binary that created just with 'make' command without 'make install' with Pinpoint.
I guess the reason is that Pinpoint has thrift libraries (libthrift-*.jar) already because Pinpoint is a maven project. 

And I added several exception handling code, too.

Thank you.

koo-taejin:
@chaerin16 thanks for you're contribution :)

PR #4151, opened 2018-05-23
fix error when no sampled in the dubbo

lanzuyou:
When the profiler.sampling.rate > 1, the DubboProviderInterceptor.before() have too many error log in the logger.warn(). Because the "final SpanRecorder recorder = trace.getSpanRecorder();" alway null when trace.canSampled() is false.


jaehong-kim:

Issue #4414, opened 2018-07-27
AgentStat.jvmGcDetailed.gcNewTime error


yjqg6666:
**What version of pinpoint are you using?**
master/v1.7.3/v1.7.2 etc.

**Describe the bug**
From /getAgentStat/jvmGcDetailed.pinpoint, the  gcNewTime always equals to gcNewCount

**Expected behavior**
gcNewTime & gcNewCount should be different.

yjqg6666:
Fixed in PR naver/pinpoint#4415.

Xylus:
@yjqg6666 Thanks for reporting and the fix!

PR #4415, opened 2018-07-27
fix AgentStat.jvmGcDetailed.gcNewTime error

yjqg6666:
Fix naver/pinpoint#4414.


Xylus:
@yjqg6666 Thanks for your contribution :+1:

PR #4427, opened 2018-07-31
Fixes hostname retrieval triggering possible DNS lookups for Thrift plugin

### LINKED ISSUE #4389: thrift plugin InetSocketAddress#getHostName issue
**What version of pinpoint are you using?**
v1.7.3

**Describe the bug**
when thrift plugin retrieve connection information, it use InetSocketAddress#getHostName method , and may block the bussiness thread. 

**Expected behavior**
In order to avoid the main process blocked, this API should not be used.

background: 
https://stackoverflow.com/questions/10420317/java-inetaddress-gethostname-taking-a-very-long-time-to-execute

From the InetAddress#getHostName() javadocs, that method will perform **a reverse hostname lookup**. So the performance of that method call depends on the performance of the network/technology stack between the JVM and the domain name server for the target host.

In brief, that method will make a system call to perform the reverse lookup (e.g. getaddrinfo(3)) and that call will be implemented by the operating system to perform the network actions required to gather the host information via the Name Server configured for your machine.

Xylus:
This solves the issue where the Thrift plugin would attempt a reverse DNS lookup when retrieving the host name if a **TSocket** was created using an address.
Utility methods are also added, which given an **InetSocketAddress**, would retrieve the host name or the address without triggering forward/reverse DNS lookup.

resolves #4389

### DIFF
--- a/bootstrap-core/src/main/java/com/navercorp/pinpoint/bootstrap/plugin/util/SocketAddressUtils.java
+++ b/bootstrap-core/src/main/java/com/navercorp/pinpoint/bootstrap/plugin/util/SocketAddressUtils.java
@@ -1,0 +1,178 @@
+/*
+ * Copyright 2018 NAVER Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.navercorp.pinpoint.bootstrap.plugin.util;
+
+import com.navercorp.pinpoint.bootstrap.logging.PLogger;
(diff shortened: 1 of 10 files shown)


bootstrap-core/src/main/java/com/navercorp/pinpoint/bootstrap/plugin/util/SocketAddressUtils.java
```diff
@@ -0,0 +1,178 @@
...
+
+    private static final PLogger LOGGER = PLoggerFactory.getLogger(SocketAddressUtils.class);
+
+    // TODO JDK 7 InetSocketAddress.getHostString() - reflection not needed once we drop JDK 6 support.
+    private static final HostStringAccessor HOST_STRING_ACCESSOR = createHostStringAccessor();
+
+    private static HostStringAccessor createHostStringAccessor() {
+        try {
+            final Method m = InetSocketAddress.class.getDeclaredMethod("getHostString");
+            m.setAccessible(true);
```
koo-taejin, line 41: getHostString() in jdk6 is a package level accessor.
So all versions of jdk1.6 should be checked that method is included.

Xylus: @koo-taejin Thanks for the reminder.
`getHostString()` is present in 1.6.0_0 and in 1.6.0_45, so it should be there for other versions of 1.6 as well.

PR #4550, opened 2018-08-27
ice plugins

songlin411:
ice plugins


songlin411:
add config 
 profile.ice.consumer.helperlist= 
or
 profile.ice.consumer.jarpath= profile.ice.consumer.jarnamepart=

PR #4634, opened 2018-09-18
add disable table for 'ApplicationStatAggre'

imperio-wxm:
Can not drop 'ApplicationStatAggre' if not disable it.


emeroad:
Thank you for reporting

PR #4650, opened 2018-09-19
Allow plugins to detect and set application type directly

### LINKED ISSUE #4485: Enforce tighter application type detection and class file transformer registration rules
#### Description
Currently, application type detection is done after all the plugins are loaded and class file transformers are registered. This makes it impossible for plugins to utilize other plugins' type detectors to make a decision whether to register their class file transformers or not.
This led to convoluted configuration options such as `profiler.tomcat.conditional.transform`, and `profiler.jboss.conditional.transform` in order to control both plugins from registering a class file transformer for the same class.
By allowing application type detectors to run during each plugin's load time (when their `setup(...)` method is invoked), plugins can take into account what the application type is and make informed decisions whether to register their class file transformers or not from the beginning.

Another configuration option `profiler.type.detect.order`. was also present to control the order in which the **ApplicationTypeDetector**s ran for cases where multiple detectors could resolve the application as their wanted type. With this change however, we can let the plugins decide whether to override, preserve, or outright throw an exception if another application type has already been detected.

Consequently, the above configuration options can be removed, no longer confusing everyone.

Xylus:
resolves #4485


Xylus:
Closing to add ordering for loading plugins.

PR #4716, opened 2018-10-09
Fix Dubbo Plugin Test IT

VictorZeng:
Fix ide compile error, dependency dubbo-plugin is runtime.


VictorZeng:
@Xylus PTAL

Xylus:
Hi @VictorZeng 
Could you squash your commits into one? It'll clean up the commit history.

Also, does the current Dubbo integration test actually cause a compilation failure?
I'm asking because it seems to be okay for both intellij and invoking maven from command.

VictorZeng:
@Xylus in my local intellij idea compilation failure... 
```xml

        <!-- plug-ins -->
        <dependency>
            <groupId>com.navercorp.pinpoint</groupId>
            <artifactId>pinpoint-plugins</artifactId>
            <type>pom</type>
            <scope>runtime</scope>
        </dependency>
```
because the plugin dependency scope is **runtime**.

VictorZeng:
@Xylus And how to squash the commits into one...?

Xylus:
Yeh, that was the intention and hence why your fix is correct. I was confused because it isn't throwing an error on my copy of intellij. hmm.,

Anyway, you can squash using interactive rebase, which in your case would be `git rebase -i HEAD~7`.
Take a look at http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html, which explains it in more detail :)

VictorZeng:
@Xylus Thank U~

PR #4721, opened 2018-10-10
[#noissue] Fix Dubbo Plugin IT

VictorZeng:
@Xylus PTAL


Xylus:
Hi @VictorZeng 
Your commits don't look like they've been squashed :(
*Merge pull request #N from naver/master* commits should be squashed and your single *fix ide compile error, plugin dependency scope is runtime* commit should be present.

Issue #4367, opened 2018-07-14
Why is there no Hbase plugin support?


VictorZeng:
**Which module is your feature request related to?**
Plugin/Hbase

**Is your feature request related to a problem?**
A clear and concise description of what the problem is.

**Describe the solution you'd like**
Hbase GET/PUT/DELETE/SCAN etc

**Describe alternatives you've considered**

**Additional context**

RoySRose:
Hello, @VictorZeng 

We hope that we can provide all the features. But unfortunately, it's not easy due to lack of time.
It would be GREAT for Pinpoint Community if you can make Hbase-plugin and share with us:)

VictorZeng:
@RoySRose I will make the Hbase-plugin

RoySRose:
@VictorZeng 
Oh! Perfect~! We are really looking forward to it~!

PR #4732, opened 2018-10-13
Add Hbase Plugin

VictorZeng:
Title: Hbase Plugin Contribution

Link: https://github.com/VictorZeng/pinpoint/tree/hbase-plugin/plugins/hbase
Target: pinpoint-hbase-plugin
Supported Version: hbase version 1.0.0+.
Description: trace hbase client and show hbase topology in the web ui.

ServiceTypes: 

- (8800, "HBASE_CLIENT")
- (8801, "HBASE_CLIENT_ADMIN") 
- (8802, "HBASE_CLIENT_TABLE")
- (8803, "HBASE_ASYNC_CLIENT")

Annotations:

-  (320, "hbase.client.params", VIEW_IN_RECORD_SET)

Configurations: 

- profiler.hbase.client.enable=true
- profiler.hbase.client.admin.enable=true
- profiler.hbase.client.table.enable=true
- profiler.hbase.client.params.enable=true


VictorZeng:
tim 20181013133105

VictorZeng:
tim 20181013133619

VictorZeng:
tim 20181013134133

VictorZeng:
fix #4367

VictorZeng:
@Xylus PTAL~

Xylus:
@VictorZeng On it!

plugins/hbase/src/main/java/com/navercorp/pinpoint/plugin/hbase/HbasePluginConstants.java
```diff
@@ -0,0 +1,73 @@
...
+ */
+public final class HbasePluginConstants {
+
+    private HbasePluginConstants() {
+    }
+
+    /**
+     * The constant HBASE.
+     */
+    public static final ServiceType HBASE = ServiceTypeFactory.of(8800, "HBASE", TERMINAL, RECORD_STATISTICS);
```
Xylus: Would this work better if renamed to something like `HBASE_CLIENT` instead of `HBASE`?
If there comes a time when we add support for hbase itself (not hbase client), then the names could get confusing.

Xylus: If you decide to rename this to `HBASE_CLIENT`, it might be a good idea to rename the variables below as well like :
```
HBASE_ADMIN -> HBASE_CLIENT_ADMIN
HBASE_TABLE -> HBASE_CLIENT_TABLE
hbase.params -> hbase.client.params
profiler.hbase.enable -> profiler.hbase.client.enable
profiler.hbase.operation.enable -> profiler.hbase.client.operation.enable
```
etc

plugins/hbase/src/main/java/com/navercorp/pinpoint/plugin/hbase/interceptor/HbaseClientMethodInterceptor.java
```diff
@@ -0,0 +1,78 @@
...
+     */
+    protected String getEndPoint(Object[] args) {
+        // call(PayloadCarryingRpcController pcrc, MethodDescriptor md,
+        //      Message param, Message returnType, User ticket, InetSocketAddress addr,
+        //      MetricsConnection.CallStats callStats)
+        if (args != null && args.length == 7) {
+
+            if (args[5] instanceof InetSocketAddress) {
+
+                return args[5].toString();
```
Xylus: We have utility method for retrieving address from **InetSocketAddress**.
Please take a look at **[InetSocketAddressUtils.getHostNameFirst(...)](https://github.com/naver/pinpoint/blob/master/bootstrap-core/src/main/java/com/navercorp/pinpoint/bootstrap/plugin/util/SocketAddressUtils.java#L100)** method.

plugins/hbase/src/main/java/com/navercorp/pinpoint/plugin/hbase/interceptor/HbaseClientMethodInterceptor.java
```diff
@@ -0,0 +1,78 @@
...
+    @Override
+    protected void doInBeforeTrace(SpanEventRecorder recorder, Object target, Object[] args) {
+        recorder.recordServiceType(HbasePluginConstants.HBASE);
+    }
+
+    @Override
+    protected void doInAfterTrace(SpanEventRecorder recorder, Object target, Object[] args, Object result, Throwable throwable) {
+        String endPoint = getEndPoint(args);
+        recorder.recordEndPoint(endPoint != null ? endPoint : "Unknown");
+        recorder.recordDestinationId(HbasePluginConstants.HBASE.getName());
```
Xylus: Just a note for the future, destination id should be a logical name of the destination node.
AFAIK, terminal nodes with the same destination id are merged into a single node in the server map, even though they have different end points. (Think of destination id == application name, end point == agent id for terminal nodes)

I'm not sure what would be a good value to use as destination id for hbase (maybe zk quorum + znode?), so we'll leave this as is for now.
If you have a good idea for this, please let me know.

plugins/hbase/src/main/java/com/navercorp/pinpoint/plugin/hbase/interceptor/HbaseTableMethodInterceptor.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    @Override
+    protected void doInBeforeTrace(SpanEventRecorder recorder, Object target, Object[] args) {
+        recorder.recordServiceType(HbasePluginConstants.HBASE_TABLE);
+    }
+
+    @Override
+    protected void doInAfterTrace(SpanEventRecorder recorder, Object target, Object[] args, Object result, Throwable throwable) {
+        String attributes = parseAttributes(args);
```
Xylus: I'd suggest adding an option to record parameters in pinpoint.config (default: false) here as well.

plugins/hbase/src/main/java/com/navercorp/pinpoint/plugin/hbase/interceptor/HbaseAdminMethodInterceptor.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    @Override
+    protected void doInBeforeTrace(SpanEventRecorder recorder, Object target, Object[] args) {
+        recorder.recordServiceType(HbasePluginConstants.HBASE_ADMIN);
+    }
+
+    @Override
+    protected void doInAfterTrace(SpanEventRecorder recorder, Object target, Object[] args, Object result, Throwable throwable) {
+        String attributes = parseAttributes(args);
```
Xylus: I'd recommend adding an option to record parameters in pinpoint.config (default: false).

plugins/hbase/src/main/java/com/navercorp/pinpoint/plugin/hbase/HbasePluginConstants.java
VictorZeng: Thank you for your advice, I'll adjust it.

plugins/hbase/src/main/java/com/navercorp/pinpoint/plugin/hbase/interceptor/HbaseClientMethodInterceptor.java
VictorZeng: Thank you for your advice, I'll adjust it.

plugins/hbase/src/main/java/com/navercorp/pinpoint/plugin/hbase/interceptor/HbaseTableMethodInterceptor.java
VictorZeng: Thank you for your advice, I'll adjust it.

plugins/hbase/src/main/java/com/navercorp/pinpoint/plugin/hbase/interceptor/HbaseAdminMethodInterceptor.java
VictorZeng: Thank you for your advice, I'll adjust it.

plugins/hbase/src/main/java/com/navercorp/pinpoint/plugin/hbase/interceptor/HbaseClientMethodInterceptor.java
VictorZeng: Thank U for your comments.
AFAIK, Hbase Client gets the real Region Server Address through the Zookeeper, then use RPC to operate(PUT/GET/SCAN/...) Region Server, so I think use the real address record end point. 

About the DestinationId, I refer to the implementation of Redis Plugin:
[recorder.recordDestinationId(RedisConstants.REDIS.getName());](https://github.com/naver/pinpoint/blob/master/plugins/redis/src/main/java/com/navercorp/pinpoint/plugin/redis/interceptor/JedisMethodInterceptor.java#L79)
What do you recommend as a DestinationId?

Xylus: Yes, you're absolutely correct, the client connects directly to region servers using the meta data stored in zookeeper. Your end point recording is correct.
I'm not sure what we should we use for destination id so we can go with `recorder.recordDestinationId(HbasePluginConstants.HBASE.getName());` for the purpose of this pull request.

If you do come up with a good value for the destination id, please let us know.

VictorZeng: OK

Xylus:
@VictorZeng 
Are you looking to add an interceptor to a method in a non-public class?
Or are you looking to get a reference (and cast) to a non-public class instance in your interceptor?

VictorZeng:
@Xylus I hope add an interceptor to a method in a non-public class.

Xylus:
@VictorZeng 
Ah, you can do it like you do a public class.
```
transformTemplate.transform("org.apache.hadoop.hbase.client.AsyncProcess", new TransformCallback() {
    ...
}
```

VictorZeng:
@Xylus And How to add an interceptor to a constructor method in a non-public class.
image

Xylus:
@VictorZeng 
For that, it would be something like 
```
transformTemplate.transform("org.apache.hadoop.hbase.client.AsyncProcess$AsyncRequestFutureImpl$SingleServerRequestRunnable", new TransformCallback() {
    @Override
    public byte[] doInTransform(Instrumentor instrumentor, ClassLoader loader, String className, Class<?> classBeingRedefined, ProtectionDomain protectionDomain, byte[] classfileBuffer) throws InstrumentException {
        InstrumentClass target = instrumentor.getInstrumentClass(loader, className, classfileBuffer);
        InstrumentMethod c = target.getConstructor("org.apache.hadoop.hbase.client.MultiAction", "int", "org.apache.hadoop.hbase.ServerName", "java.util.Set");
        if (c != null) {
            // Add your interceptor
        }
        ...
        return target.toByteCode();
}
```

Xylus:
@VictorZeng 
Ah sorry, non-static inner class constructors have a reference to their outer class instance added as their first parameter by the compiler.
Try this and let me know if it works.
```
        InstrumentMethod c = target.getConstructor("org.apache.hadoop.hbase.client.AsyncProcess$AsyncRequestFutureImpl", "org.apache.hadoop.hbase.client.MultiAction", "int", "org.apache.hadoop.hbase.ServerName", "java.util.Set");
```

VictorZeng:
@Xylus It's works. Thank U so much.
image

VictorZeng:
@Xylus Hi, Do I need to pull a new request?

VictorZeng:
@Xylus There seems to be something wrong with Github...

VictorZeng:
Hi, Xylus

There seems to be something wrong with Github...

发件人: HyunGil Jeong <notifications@github.com>
发送时间: 2018年10月22日 13:09
收件人: naver/pinpoint <pinpoint@noreply.github.com>
抄送: 业余布道师 <Victor.Zxy@outlook.com>; Mention <mention@noreply.github.com>
主题: Re: [naver/pinpoint] Add Hbase Plugin (#4732)


@Xylus approved this pull request.

lgtm, good job 👍
Could you fix the conflict for plugins/bom/pom.xml so that I merge?

Xylus:
@VictorZeng 
I think the last 2 commits (cc638d1c3b0d09d068c2720ec0167af9efdee388, 0edf213642c1ed80bc75b05fbb2f9b9b7dec92ac) broke the PR.
You could reset these 2 commits (from your local hbase-plugin branch) and push -f to get rid of them.

VictorZeng:
@Xylus why Merging is blocked?

Xylus:
@VictorZeng It keeps saying that the branch cannot be rebased due to conflicts. Hmm...
Could you open a new PR and see if github works itself out?

VictorZeng:
@Xylus OK

PR #4744, opened 2018-10-16
Mongo fix

RoySRose:


RoySRose:
remove MongoDB code from common
enhance performance by reprocessing bson map

RoySRose:
removed trim

PR #4804, opened 2018-10-26
Kafka 2.0.0+ support

kwangil-ha:
Kafka 2.0.0+ support


koo-taejin:
hi @kwangil-ha 

Currently, I am checking another  pull request associated with kafka.
Let's take a look at it after that.

Thank you for you're contribution

PR #4813, opened 2018-10-29
[#noissue] Optimize cxf plugin.

VictorZeng:
@Xylus PTAL...


Xylus:
@VictorZeng 
Thank you for all your work!

Removing reflection from interceptors is usually a good idea, but it can be limiting at times.
For example, `LoggingMessage.getId()` method seems to have been introduced in apache cxf 3.0.0, so if users are running apache cxf 2.x, the interceptor will blow up and fail.
Using reflection, you will be explicitly checking if the method exists or not and will be catching/handling **NoSuchMethodException** proactively so the side effect is usually reduced.

Please let me know what your thoughts are regarding this.

VictorZeng:
Hi  @Xylus, AFAK, [`LoggingMessage`](https://github.com/apache/cxf/blob/cxf-2.6.0/api/src/main/java/org/apache/cxf/interceptor/LoggingMessage.java) since Apache CXF 2.6.0 version, It has `getId()` method. 
 
CXF Log Interceptor  `LoggingInInterceptor`  and `LoggingOutInterceptor` since 2.6.0 version too.

So, I think It doesn't matter,

VictorZeng:
@Xylus And `LoggingInInterceptor.formatLoggingMessage` and `LoggingOutInterceptor.formatLoggingMessage` since 2.6.2 version. I think I should add comments to the config file for describe this.

Xylus:
@VictorZeng 
Thanks for checking, are you sure about the `getId()` method for **LoggingMessage**?
It looks like it got added in [2.7.12](https://github.com/apache/cxf/blob/cxf-2.7.12/api/src/main/java/org/apache/cxf/interceptor/LoggingMessage.java#L53) / [2.6.15](https://github.com/apache/cxf/blob/cxf-2.6.15/api/src/main/java/org/apache/cxf/interceptor/LoggingMessage.java#L53), since [2.7.11](https://github.com/apache/cxf/blob/cxf-2.7.11/api/src/main/java/org/apache/cxf/interceptor/LoggingMessage.java) and [2.6.14](https://github.com/apache/cxf/blob/cxf-2.6.14/api/src/main/java/org/apache/cxf/interceptor/LoggingMessage.java) don't seem to have the method.

Either way, if you think 2.7.11 / 2.6.14 are way too old and not many people would use them anymore, then we can go ahead with the PR.
It would also be a good idea to add a *README.md* file to *plugins/cxf* directory and list the supported versions.
Let me know what you think.
Thanks!

VictorZeng:
@Xylus I'm sorry I misread it, you're right, I think remove the `cxf.log.id` is better, what you think?

Xylus:
@VictorZeng 
I don't have any experience with apache cxf so it will be better if you decide on it :)
If you think recording `cxf.log.id` is important (if it would be useful for users if they can check the id) then please feel free to include it!
If you don't think `cxf.log.id` is very important, you can drop it.

VictorZeng:
@Xylus Thank U so mush, This `id `is CXF to associate Request and Response, I think it's not required in a Trace. I will remove it.

plugins/cxf/src/main/java/com/navercorp/pinpoint/plugin/cxf/interceptor/CxfLoggingMessageMethodInterceptor.java
```diff
@@ -49,6 +50,7 @@ protected CxfLoggingMessageMethodInterceptor(TraceContext traceContext, MethodDe
      * @param methodName the method name
      * @return the attribute
      */
+    @Deprecated
     protected <T> T getAttribute(Object object, String methodName) {
```
Xylus: I think we can safely remove this method instead of deprecating since the method has not been part of any official release yet.

VictorZeng: OK

VictorZeng:
@Xylus I will squash the commits into one PR

VictorZeng:
@Xylus By the way, when do you plan to release version 1.8.1 ?

PR #4814, opened 2018-10-29
[#noissue] Optimize druid plugin.

VictorZeng:
@Xylus PTAL...


VictorZeng:
`DruidDataSource.getUrl()` 
`DruidDataSource.getActiveCount()`
`DruidDataSource.getMaxActive()`
These methods have existed since the original version of the Druid.
So. I think remove the reflection from interceptors.

PR #4875, opened 2018-11-09
ElasticsearchBBoss Client Plugin Contribution

yin-bp:
Title: Elasticsearch Client Plugin Contribution

Link: https://github.com/yin-bp/pinpoint/tree/master/plugins/elasticsearch-bboss
Target: Elasticsearch Rest Client BBoss
Supported Version: All Elasticsearch version,BBoss version 5.1.9,pinpoint 1.8.1
Description: This agent plugin is worked with elasticsearch rest client BBoss,BBoss is a best Java Highlevel Rest client for ElasticSearch.

ServiceTypes: [1971, "ElasticsearchBBoss"],[1972, "ElasticsearchBBossEvent"],[9977, "ElasticsearchBBossExecutor"]
Annotations: [971, "es.args"],[972, "es.url"],[973, "es.dsl"],[974, "es.action"],[975, "es.responseHandle"],[976, "es.version"]
Configurations: List of configuration keys and description the plugin adds.

profiler.elasticsearchbboss.enabled=true //enable or disable this plugin, default true.
profiler.elasticsearchbboss.recordResult=false //record or disable record method return value ,default false.
profiler.elasticsearchbboss.recordArgs=true //record or disable record method args,default true.
profiler.elasticsearchbboss.recordDsl=true //record or disable record dsl,default true.
profiler.elasticsearchbboss.recordESVersion=true //record or disable record ES Version infomation,default true.
profiler.elasticsearchbboss.recordResponseHandlerClass=false //record or disable record response handle class ,default false.


plugins/elasticsearch-bboss/src/main/java/com/navercorp/pinpoint/plugin/elasticsearchbboss/interceptor/ElasticsearchBaseOperationInterceptor.java
```diff
@@ -0,0 +1,237 @@
...
+				else{
+					convertArray(  arg,  builder);
+				}
+			}
+			return builder.toString();
+		}
+		return null;
+	}
+
+	public void convertArray(Object arg,StringBuilder builder){
```
RoySRose: Hello,  @yin-bp 
Currently the process of logic is, convertParams() -> convertArray() when there is an array inside a parameter.
What happens if there is an Array inside an Array? is this case never going to happen?

plugins/elasticsearch-bboss/src/main/java/com/navercorp/pinpoint/plugin/elasticsearchbboss/ElasticsearchCustomMethodFilter.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+public class ElasticsearchCustomMethodFilter implements MethodFilter {
+
+	public ElasticsearchCustomMethodFilter( ) {
+
+
+	}
+
+	@Override
+	public boolean accept(InstrumentMethod method) {
```
RoySRose, line 33: @yin-bp 

Pinpoint, usually, recommend specifying the method name due to the update of ElasticBBoss library.
Since any error can occur when the version is updated. If methods are specified instead of excluding certain method. It will be much safer to the version updates.

[Here is an example of it, check `getMethodlistR3_7_x()` ](https://github.com/naver/pinpoint/blob/master/plugins/mongodb/src/main/java/com/navercorp/pinpoint/plugin/mongo/MongoPlugin.java)

plugins/elasticsearch-bboss/src/main/java/com/navercorp/pinpoint/plugin/elasticsearchbboss/interceptor/ElasticsearchExecutorOperationInterceptor.java
```diff
@@ -0,0 +1,230 @@
...
+        } else {
+//            recorder.recordApi(getMethodDescriptor());
+        }
+
+        if(recordResult){
+        	recorder.recordAttribute(AnnotationKey.RETURN_DATA,result);
+		}
+    }
+
+    private void recordAttributes(SpanEventRecorder recorder, MethodDescriptor methodDescriptor,Object[] args){
```
RoySRose, line 106: there are duplicate or similar codes in two record Attributes. What do you think of combining the ones that are possible?

plugins/elasticsearch-bboss/src/main/java/com/navercorp/pinpoint/plugin/elasticsearchbboss/ElasticsearchPlugin.java
```diff
@@ -0,0 +1,151 @@
...
+	public static final AnnotationKey ARGS_URL_ANNOTATION_KEY = AnnotationKeyFactory.of(972, "es.url", AnnotationKeyProperty.VIEW_IN_RECORD_SET);
+	public static final AnnotationKey ARGS_DSL_ANNOTATION_KEY = AnnotationKeyFactory.of(973, "es.dsl", AnnotationKeyProperty.VIEW_IN_RECORD_SET);
+	public static final AnnotationKey ARGS_ACTION_ANNOTATION_KEY = AnnotationKeyFactory.of(974, "es.action", AnnotationKeyProperty.VIEW_IN_RECORD_SET);
+	public static final AnnotationKey ARGS_RESPONSEHANDLE_ANNOTATION_KEY = AnnotationKeyFactory.of(975, "es.responseHandle", AnnotationKeyProperty.VIEW_IN_RECORD_SET);
+	public static final AnnotationKey ARGS_VERSION_ANNOTATION_KEY = AnnotationKeyFactory.of(976, "es.version", AnnotationKeyProperty.VIEW_IN_RECORD_SET);
+	public static final ServiceType ELASTICSEARCH_EXECUTOR = ServiceTypeFactory.of(9977, "ElasticsearchBBossExecutor");
+	private static final String ELASTICSEARCH_SCOPE = "ElasticsearchBBoss_SCOPE";
+	private static final String ELASTICSEARCH_EXECUTOR_SCOPE = "ElasticsearchBBossExecutor_SCOPE";
+	private static final String[] clazzInterceptors = new String[]{
```
RoySRose, line 55: I'm not really sure about this since I didn't go through all the methods. But is it really necessary to instrument both
"org.frameworkset.elasticsearch.client.ConfigRestClientUtil",
"org.frameworkset.elasticsearch.client.RestClientUtil"
classes?

Because if I look into the `ConfigRestClientUtil` class . It is the subclass of `RestClientUtil` and most of the method eventually calls `RestClientUtil`'s method
```java
public class ConfigRestClientUtil extends RestClientUtil {
```
For example, *count* method eventually calls super.count()
so, I think you should check if it is possible for the plug-in if only `RestClientUtil` class is instrumented.
```java
       public long count(String index,String templateName)  throws ElasticSearchException{
		return  super.count(index, ESTemplateHelper.evalTemplate(esUtil,templateName, (Object) null));
	}
	public long count(String index,String templateName,Map params)  throws ElasticSearchException{
		return super.count(index, ESTemplateHelper.evalTemplate(esUtil,templateName, params));
	}
	public long count(String index,String templateName,Object params)  throws ElasticSearchException{
		return super.count(index, ESTemplateHelper.evalTemplate(esUtil,templateName, params));
	}
```

RoySRose:
@yin-bp, Most of all. 

I'm studying BBossES Client to get to know about them.
As shown in this [link on bboss-elastic](https://www.oschina.net/p/bboss-elastic)

*Bboss elasticsearch is a set of o/r mapping high-performance development library based on query dsl syntax operation and access to distributed search engine elasticsearch. The bottom layer is based on es restful api.*

If the bottom layer of Bboss elasticsearch is ES restful api. You should check if there can be any conflicts between Bboss elasticsearch plugin and HTTP4 plugin(since the bottom layer of ES resful api is HTTP4). I'm also trying to check this issue. But having some hard time on setting up the env.

RoySRose:
@yin-bp 

A few tracked data is duplicated and looks unnecessary.

image

plugins/elasticsearch-bboss/src/main/java/com/navercorp/pinpoint/plugin/elasticsearchbboss/interceptor/ElasticsearchBaseOperationInterceptor.java
yin-bp: Hello,@RoySRose
The API method with an array of parameters exists, For example, the mget method has a variable-length array parameter ids
```java
/**
	 * https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-multi-get.html
	 * @param index _mget
	 *             test/_mget
	 *             test/type/_mget
	 *             test/type/_mget?stored_fields=field1,field2
	 *             _mget?routing=key1
	 * @param indexType
	 * @param ids
	 * @return
	 * @throws ElasticSearchException
	 */
	public String mgetDocuments(String index,String indexType,Object ... ids)  throws ElasticSearchException;
```

plugins/elasticsearch-bboss/src/main/java/com/navercorp/pinpoint/plugin/elasticsearchbboss/ElasticsearchCustomMethodFilter.java
yin-bp: Hi,@RoySRose
ElasticBBoss library API is very stable, the pinpoint interceptor filter does not need to correspond to its version number.

plugins/elasticsearch-bboss/src/main/java/com/navercorp/pinpoint/plugin/elasticsearchbboss/ElasticsearchPlugin.java
yin-bp: > ConfigRestClientUtil

plugins/elasticsearch-bboss/src/main/java/com/navercorp/pinpoint/plugin/elasticsearchbboss/interceptor/ElasticsearchExecutorOperationInterceptor.java
yin-bp: Hi,@RoySRose
As you suggest, duplicate code has been merged.

plugins/elasticsearch-bboss/src/main/java/com/navercorp/pinpoint/plugin/elasticsearchbboss/ElasticsearchPlugin.java
yin-bp: RestClientUtil directly executes the DSL defined in the code.ConfigRestClientUtil gets the DSL defined in the configuration file by the DSL name and executes it. Therefore, it is necessary to collect their call parameter and component information separately.

RestClientUtil example:
```java
ClientInterface clientUtil = ElasticSearchHelper.getRestClientUtil();//define an instanceof RestClientUtil,It's single instance, multithreaded secure.
List<Map> json = clientUtil.sql(Map.class,"{\"query\": \"SELECT * FROM demo\"}");
```

ConfigRestClientUtil example:
```java
ClientInterface clientUtil = ElasticSearchHelper.getConfigRestClientUtil("esmapper/sql.xml");//define an instanceof ConfigRestClientUtil,It's single instance, multithreaded secure.
Map params = new HashMap();
params.put("channelId",1);
List<Map> json = clientUtil.sql(Map.class,"sqlQuery",params);
```

the dsl config file:sql.xml ,We can define a lot of DSLs in the configuration file
```xml
<properties>
    <!--
        sql query
    -->
    <property name="sqlQuery">
        <![CDATA[
         {"query": "SELECT * FROM dbclobdemo where channelId=#[channelId]"}
        ]]>
    </property>
</properties>
```

yin-bp:
> @yin-bp
> 
> A few tracked data is duplicated and looks unnecessary.
> 
> image

How to set up elasticsearch bboss to disable http4 plugin？

yin-bp:
move to pr #4875

Issue #4737, opened 2018-10-15
MongoDB Plugin clean up & enhance performance


RoySRose:
MongoDB Plugin clean up & enhance performance

PR #4917, opened 2018-11-27
[#4737] match format of JsonWriter

RoySRose:
matching output format to default format of JsonWriter(JsonMode.STRICT)


RoySRose:
abbreviation
image

Issue #4939, opened 2018-12-03
Supports spring-kafka via option within the kafka plugin. 


koo-taejin:
### description 
Supports spring-kafka via option within the kafka plugin. 

#### Option
'profiler.springskafka.consumer.enable=true/false'

#### Trace policy 
spring-kafka  also follows the previous trace policy. 
Per case: remote trace support, 
batch case : remote trace is not supported, and trace the starting point.

#### Change virtual node expression
as-is : topic=topicName 
to-be :  network address 

Reason : Kafka node can have multiple topic, and some case of listener  is often used in combination with multiple topic. Multiple topics make it difficult to represent virtual nodes.
So, rename the virtual node to address values.

PR #4940, opened 2018-12-03
[#4939] Supports spring-kafka via option within the kafka plugin

koo-taejin:
Supports spring-kafka via option within the kafka plugin

### DIFF
--- a/agent/src/main/resources-local/pinpoint.config
+++ b/agent/src/main/resources-local/pinpoint.config
@@ -874,6 +874,8 @@ profiler.akka.http.transform.targetparameter=scala.Function1,scala.Function1,akk
 ###########################################################
 profiler.kafka.producer.enable=false
 profiler.kafka.consumer.enable=false
+# Setting when using spring-kafka (In this case, you can leave profiler.kafka.consumer.entryPoint option to empty.)
+profiler.springkafka.consumer.enable=false
 // you must set target that handles ConsumerRecord or ConsumerRecords(Remote Trace feature is not enabled.) as a argument for remote trace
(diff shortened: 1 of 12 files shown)


agent/src/main/resources-local/pinpoint.config
```diff
@@ -874,6 +874,8 @@ profiler.akka.http.transform.targetparameter=scala.Function1,scala.Function1,akk
 ###########################################################
 profiler.kafka.producer.enable=false
 profiler.kafka.consumer.enable=false
+# Setting when using spring-kafka (In this case, you can leave propeller.kafka.conscher.entryPoint option to empty.)
```
Xylus: `propeller.kafka.conscher.entryPoint` typo?

agent/src/main/resources-release/pinpoint.config
```diff
@@ -865,6 +865,8 @@ profiler.akka.http.transform.targetparameter=scala.Function1,scala.Function1,akk
 ###########################################################
 profiler.kafka.producer.enable=false
 profiler.kafka.consumer.enable=false
+# Setting when using spring-kafka (In this case, you can leave propeller.kafka.conscher.entryPoint option to empty.)
```
Xylus: typo as well I think.

agent/src/main/resources-local/pinpoint.config
koo-taejin: Thanks for you're review. I fixed it :)

agent/src/main/resources-release/pinpoint.config
koo-taejin: Thanks for you're review. I fixed it :)

PR #4998, opened 2018-12-13
Support latest version of Openwhisk

upgle:
## Note

Currently, Openwhisk plugin does not work with the openwhisk's latest version.
because there have been following major changes in the code recently. 

- scala version is updated from 2.11 to 2.12
- package name is changed from `whisk` to `org.apache.openwhisk`

This plugin has been modified and tested to work with the latest version.

## Changes
- Scala version is updated from 2.11 to 2.12
  - need to use `scala.Function0` to get a log message instead of `scala.runtime.AbstractFunction0`
- a package name is changed from `whisk` to `org.apache.openwhisk`
- Added parameter config `profiler.openwhisk.transform.targetparameter` to accommodate upstream changes more flexible.


upgle:
@koo-taejin  
I'm so sorry to keep you waiting. the conflict has been resolved and i've finished testing this feature in the latest master version. 
Could you please review it again?

koo-taejin:
@upgle thanks for you're contribution

Issue #5048, opened 2018-12-26
Group the request and report the call-chain by the request group. 


xiaoshao:
**Which module is your feature request related to?**
ALL

**Is your feature request related to a problem?**
NO

**Describe the solution you'd like**
I found that the requests are grouped by the business in some apm product for example the dynatrace.

For example when the target app is based on spring mvc.
The requests are grouped by the handler methods which has RequestMapping annotations.
And report the call-chain data by the group.

**Describe alternatives you've considered**

**Additional context**

xiaoshao:
hi guys, I am a java developer and I am interested in pinpoint. If pinpoint plan to accept my suggestion, I can contribute my code. thanks guys.  looking forward for your final decision.

emeroad:
@xiaoshao Hi
I do not know what call-chain means.
Does this feature mean URL statistics?

Example Screenshot
https://github.com/slanatech/swagger-stats/raw/master/screenshots/methods.png?raw=true
![image](https://user-images.githubusercontent.com/7564547/50500675-41fd4400-0a96-11e9-98e7-d55d05cff10a.png)

xiaoshao:
sorry, we named the call stack in pinpoint with call-chain.

Yes, I suggest to group the call stack with URL statistics.

For example there is some code in the Spring MVC as following.

```
@Controller
public class SecondController {

    @GetMapping("/address/{name}")
    @ResponseBody
    public List<String> address(@PathVariable("name") String name) {
        List<String> address = new ArrayList<>();
        address.add("first");
        address.add("second");
        return address;
    }
}
``` 

We can group the callstack all the request which is handled by the annotation of the method `address`, even more we can report the callstack by this to avoid that some handler is missed.

what do you think about it.

emeroad:
We welcome your suggestions

I agree with your idea.
We have plans to aggregate statistical data into the RequestMapping annotations, 
but we do not have enough time to develop.

Below is our idea.
1. Use RequestMapping annotation.
    - When using real url, it is difficult to aggregate rest style url
       - eg: /book/{id}/
2. Unable to handle annotation in bytecode api
   - We will add API.
   - If the annotation data can be extracted to the runtime, API addition is unnecessary
3. Aggregation plan
    1. agent memory
    2. collector
      - Use the `Table.incrementColumnValue()` API of hbase
    3. flink
    **Not determined**
  4. Data type
    1. Histogram
        - Counts 1s, 3s, 5s
     2. total reponse time
    **Not determined**

[...]

PR #5054, opened 2018-12-29
for the issue https://github.com/naver/pinpoint/issues/5048

xiaoshao:
hi @emeroad, I did not finish all the work about the transaction idea.

Just want to you have a look. I can change it earlier if you have any concerns.

Looking forward for your suggestions.

thanks very much.

### DIFF
--- a/bootstrap-core/src/main/java/com/navercorp/pinpoint/bootstrap/context/Header.java
+++ b/bootstrap-core/src/main/java/com/navercorp/pinpoint/bootstrap/context/Header.java
@@ -28,6 +28,7 @@ public enum Header {
     HTTP_PARENT_SPAN_ID("Pinpoint-pSpanID"),
     HTTP_SAMPLED("Pinpoint-Sampled"),
     HTTP_FLAGS("Pinpoint-Flags"),
+    HTTP_TRANSACTION_TYPE("Pinpoint-TransactionType"),
     HTTP_PARENT_APPLICATION_NAME("Pinpoint-pAppName"),
     HTTP_PARENT_APPLICATION_TYPE("Pinpoint-pAppType"),
     HTTP_PARENT_APPLICATION_NAMESPACE("Pinpoint-pAppNamespace"),
--- a/bootstrap-core/src/main/java/com/navercorp/pinpoint/bootstrap/context/TraceContext.java
+++ b/bootstrap-core/src/main/java/com/navercorp/pinpoint/bootstrap/context/TraceContext.java
@@ -44,6 +44,8 @@ public interface TraceContext {
 
     Trace newTraceObject();
 
+    Trace newTraceObject(String transactionType);
+
     /**
      * internal experimental api
      */
@@ -90,6 +92,7 @@ public interface TraceContext {
     boolean cacheSql(ParsingResult parsingResult);
 
     TraceId createTraceId(String transactionId, long parentSpanId, long spanId, short flags);
+    TraceId createTraceId(String transactionId, long parentSpanId, long spanId, short flags, String transactionType);
 
     Trace disableSampling();
 
--- a/bootstrap-core/src/main/java/com/navercorp/pinpoint/bootstrap/context/TraceContext.java
+++ b/bootstrap-core/src/main/java/com/navercorp/pinpoint/bootstrap/context/TraceContext.java
@@ -17,6 +17,7 @@
 package com.navercorp.pinpoint.bootstrap.context;
 
 import com.navercorp.pinpoint.bootstrap.config.ProfilerConfig;
+import com.navercorp.pinpoint.bootstrap.context.transaction.IMappingRegistry;
 import com.navercorp.pinpoint.bootstrap.plugin.jdbc.JdbcContext;
 import com.navercorp.pinpoint.common.annotations.InterfaceAudience;
 import com.navercorp.pinpoint.common.annotations.InterfaceStability;
@@ -97,5 +98,7 @@ public interface TraceContext {
     ServerMetaDataHolder getServerMetaDataHolder();
 
     JdbcContext getJdbcContext();
+
+    IMappingRegistry getMappingRegistry();
 
 }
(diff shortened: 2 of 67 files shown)


minwoo-jung:
@xiaoshao 

thank you for good idea^^
I will review this code.

xiaoshao:
@minwoo-jung  Did you review my code? Do you have any suggestion about it?

xiaoshao:
@minwoo-jung I think I have finished the http protocol.  Plz let me know it if you have any concerns about this.

Looking forward for you reply.

minwoo-jung:
@xiaoshao
I finished review. I will give a few feedback tomorrow.

minwoo-jung:
@xiaoshao

I think your idea is good. Also we thought url statistics is needed.


**1** Does agent works fine? When I executed agent, exception occurred.

**2** I think url pattern data should be acquired from inside the spring then directly matching url pattern in RequestTraceReader.

```
RequestTraceReader.java

private Trace newTrace(T request) {
    String txType = traceContext.getMappingRegistry().match(requestAdaptor.getRpcName(request), requestAdaptor.getMethod(request)).getTransactionType();
```

You should find matching request and requestmapping annotation's propteries at controller method in spring.
[This commit(HandleMatchInterceptor.java)](https://github.com/minwoo-jung/pinpoint/commit/b1fd045fbaf733bdcab33c3993d6e991f55ce855#diff-cc4804b481b73012f6077c3ae2db60cb) may help you analysis a internal spring. 


**3** I think url pattern doesn't need to include TraceId Object. 
It seem to be only included in the span object. 
Can you explain the reason why?

**4** I don't know why there is transactiontype http header.
Can you explain the reason why?

**5** Sampling request is only collected url pattern data. So it would be good to remove SamplingTransactionSampler.
How did you thinks of this?

**6** It seems better to change transactionType word like url Pattern.

**7** I recommend use logger to logging then system.out.print(~).

xiaoshao:
@minwoo-jung thanks for your review. I will consider you concerns one by one and then reply it.

xiaoshao:
@minwoo-jung It works now, I think. Plz have a look when you have time.

minwoo-jung:
1. There is a class not found ex, I have fixed that, I will commit the code later.
> good!^^

2. Do you mean that we should try to get the uri pattern after request.setAttribute(HandlerMapping.PATH_WITHIN_HANDLER_MAPPING_ATTRIBUTE, lookupPath); which is in the class AbstractHandlerMethodMapping ?
==> I read the code for find the uri pattern in spring. I think the way we are using is good and I added a configuration item for the uri pattern. User can add the uri pattern by themselves when they are not satisfy with the uri pattern.

> I think We should get url pattern by spring internal logic.
> Because we must consider performance and matching load while get url pattern.
> Therefore I recommend to get url pattern without duplicating match operation.
> I want to think about adding custom configuration after implement is complete.

3. This is just the first step, I added it in the traceId object and span.
    we can analyze the call stack after grouping by uri pattern
    Analyze the response time after grouping by uri pattern and so on.

> I think You want to classify request by url pattern or other factors in call stack view. I understand your intention. 
> But I think url pattern included only span is passed collector server. And then url pattern is extract from span and data key combined with stat and transactionId save to hbase. 
> Did you add code to inserted url pattern in span? I can't find code.

4. I add it to http header because I want to make the uri pattern is consist in the trace.

> I don't know why need to http header. 
> When trace is passed next node I think that url pattern data of before node is used in next node is wrong, new url pattern data must be set from next node.



5. SampingTransactionSampler for report the span reasonable, currently we just report the span on some rate.
we will missed some handler when the request for the handler is very lower frequency.
In this way, We will not miss it.

> We have a plan to create light span data for unsamping request. light span containes stat data(ex. url pattern, caller data etc).
> Then we can collect stat data for unsampled reqeust. we will calculate url stat for all request.
> First of all we consider sampled request. After that function is implemented, url stat for unsampled request add to url stat.
> Let's use the default sampler class.


In addition, 
I wish that url pattern contained from span is only url pattern id, not raw url pattren data.
For example, the sql statement included in SqlMetaData is delivered to collector server. span contained only sqlId.
I recommend that you review business logic for sql.

Finally, It seems like you've done a lot of thought. So I want to help you out too.

xiaoshao:
2,  We copied some  spring code for matching url pattern and I am reusing it.
I don't think it will cost a lot for getting url pattern, because there are less than 50 url pattern in the most micro services.

3.  No, I  want to finish the collect job firstly and then focus on it

4.   In my mind, It is not enough that the trace ID is same for one request. The response time for a request, we missed the important information that is from what service.
For example, there is two service (A, B) in an application.
 URL( /hello ) --> A (In a service send URL /call )--> B
 URL ( /call ) --> B
We could not analyze the response time for the B service group by request from outside or A service.

 5. Great, Could you share more about the light span data with me.  thanks

It is a good idea for report the url pattern with ID and report url pattern meta.

Thanks for your reply.  It is very kind of you.

xiaoshao:
For add the url pattern to the http header, I compared a lot APM product for example dynatrace and tingyun.

xiaoshao:
@minwoo-jung  Do you have concerns about my answer?

minwoo-jung:
We copied some spring code for matching url pattern and I am reusing it.
I don't think it will cost a lot for getting url pattern, because there are less than 50 url pattern in the most micro services.


> I think it is better to use data that already exists. 
> We should avoid duplication of logic as much as possible. We are working hard on this.
> I think it's enough for this.
> 
> To add further explanation, The logic you have developed check all of the bundles of the pattern for per every request.
> This is not a low cost. In particular, pattern matching requires a lot of cost considerations because logic must be check/look the contents of the url data.
> Finally, agent's target(user application) is not just about appplication that there are less than 50 url pattern in the most micro services.
> User Application that have more than a few hundred url patterns are also targeted.


We have to consider about all situation and condition as much as possible.^^

===============================================


In my mind, It is not enough that the trace ID is same for one request. The response time for a request, we missed the important information that is from what service.
For example, there is two service (A, B) in an application.
URL( /hello ) --> A (In a service send URL /call )--> B
URL ( /call ) --> B
We could not analyze the response time for the B service group by request from outside or A service.

> I recommend you look at traceId and spanId.
> I do not understand why url pattern of the before service(ex : A service) need to calculate the response time.
> 
> **We will have to make more definite decisions about which data to display by collecting url statistical data.
That is why we will be able to clarify specs and develop them.**
> I want you to tell me exactly what you want to page and the data you want to show.
> Therefore we agree that pinpoint need url statistics, but we need to specify what information we want to show.
> Please explain what you think about that.


Great, Could you share more about the light span data with me. thanks

> I think that is all that I am currently thinking. I will consider more later.

**We think this is a very good feature. Although it is important, the urgency is not very high, so it will be difficult to develop quickly.**
**Let's think for a long time and make good progress.**

PR #5097, opened 2019-01-07
add log4j2 plugin

licoco:
add log4j2 plugin


licoco:
Close, please. Repeat.

PR #5128, opened 2019-01-11
Overhaul pinpoint plugin loading

### LINKED ISSUE #4485: Enforce tighter application type detection and class file transformer registration rules

### LINKED ISSUE #4730: Allow plugin ids to be set inside plugin jars' manifest file
#### Description
There is no simple way to identify plugin jars at runtime apart from looking at it's filename (which can be renamed at will).
Adding an entry for plugin ids (in the form of maven's `groupId:artifactId`) in each plugin jar's manifest would allow us to identify the plugins.
We will then be able to utilize this id for agent configuration.

### LINKED ISSUE #5053: Allow plugins to be compiled with specific jdk versions
#### Description
Currently, plugins have to be compiled with jdk 6, the minimum supported java version by pinpoint (java 6).
This comes with a major disadvantage that we cannot work with features requiring compile-time dependency with higher jdk versions.
Specifying jdk version for plugin compilation is simple - just specify them in plugins' pom file.

However, things blow up in run time as pinpoint agent, collector and web, all have dependency to plugins and load plugin classes using service loader.
Loading a plugin compiled with jdk 9 from an agent running on java 6 will throw a version error, as will loading it from collector/web running on java 8.

Solving this for the agent is simple. We can check the plugin's compiler version and skip loading the plugin altogether if the current java version is lower.
For collector and web, we cannot do this as they both need trace metadata provided by the plugins via **TraceMetadataProvider** implementations, which currently require them to load the plugins.

To get around this issue, our plan is to have the plugins define trace metadata and have them registered to collector/web without having to load any classes in the plugins like below.
Have plugins define service types, annotation keys and matchers in a yaml file under `META-INF/type-provider.yml`.
```
serviceTypes:
    - code: 8310
      name: 'ACTIVEMQ_CLIENT'
      property:
          queue: true
          recordStatistics: true
      matcher:
          type: 'exact'
          code: 100 # message.queue.url
    - code: 8311
      name: 'ACTIVEMQ_CLIENT_INTERNAL'
      desc: 'ACTIVEMQ_CLIENT'
      matcher:
          type: 'args'

annotationKeys:
    - code: 101
      name: 'activemq.broker.address'
      property:
          viewInRecordSet: true
    - code: 102
      name: 'activemq.message'
      property:
          viewInRecordSet: true
```
Pinpoint collector and web will then simply parse this file and register service types, and annotation keys/matchers without using service loader to load the plugin's **TraceMetadataProvider** implementation.
Plugins will be able to obtain service type and annotation key instances created via the yml instead of defining them directly in their code, for example:
```
// ActiveMQClientConstants.java
// Before
public static final ServiceType ACTIVEMQ_CLIENT = ServiceTypeFactory.of(8310, "ACTIVEMQ_CLIENT", QUEUE, RECORD_STATISTICS);
// After
public static final ServiceType ACTIVEMQ_CLIENT = ServiceTypeProvider.getByName("ACTIVEMQ_CLIENT");
```

PS 1. *It would be great if we can have schema defined for type-provider.yml and be able to validate them.*

PS 2. *Creating plugins as multi-release jars were looked into as well. While it worked well, we dropped it as maven/IDE support for multi-release jars are not quite there yet and the cleanest solution required each plugins to be a multi-module project. Multi-release jars should definitely be looked into again when maven/IDE support improves in the future.*

Xylus:
This reorganizes how pinpoint plugins are packaged and loaded into agent, collector, and web.
Plugins may now be compiled using different jdk versions and not have class version errors on load time.
Pinpoint plugin id and java version used to compile the plugin have also been added to plugin's MANIFEST file to support this, as well as the ability to define service types and annotation keys using yaml files.
resolves #4730 
resolves #5053 

Additionally, this allows plugins to detect application types during their setup time, and to use detectors from different plugins - allowing plugins to make better decisions on detecting application type.
resolves #4485


Xylus:
Closing to rebase after another commit.

PR #5196, opened 2019-01-28
add log4j2 plugin

licoco:
add log4j2 plugin


licoco:
@minwoo-jung  

CLA  It has been signed.

minwoo-jung:
Hi @licoco.
I'm sorry about answering late. We was on holiday last week.
I have reviewed for PR. I will give you few feedbacks.

1. you'd better not use org.slf4j.MDC in LoggingEventOfLog4j2Interceptor class. Because There is dependency on slf4j library.
I think it would be better to use org.apache.logging.log4j.ThreadContext.
Please ref this (link)[https://www.baeldung.com/mdc-in-log4j-2-logback]


2. You'd better change target class for profiling. I haven't done much analysis, but I recommend profiling Log4jLogEvent as a target rather than ThreadContext.

3. So that plugin can be applied to more versions, Log4j2IT should be tested from version 2.1 or later. 
You will need to look at the structure of the classes since version 2.1.

4. I recommend git commend `rebase --interactive`  and` squash` to combine multiple commits into one.

licoco:
to  https://github.com/naver/pinpoint/pull/5246

PR #5402, opened 2019-03-25
support log4j2 plugin

licoco:


licoco:
@minwoo-jung

minwoo-jung:
Is the test success?

After reviewing the code in its entirety, it would be better to send a PR request.

Issue #5411, opened 2019-03-26
Dis/Enable Tracing embedded library


RoySRose:
Support dis/enabling function for tracing embedded library.

This feature is needed to support, libraries that have an embedded library.
such as
ElasticSearch, ElasticSearch-BBoss.
(java client that embeds http4)

PR #5413, opened 2019-03-26
[#5411] support option for tracing embedded library

RoySRose:
@emeroad 

What do you think about adding a variable in `DefaultFrameAttachment` class

such as
```
private ServiceType disabledLibrary
```


emeroad:
I think it's better to use `DefaultTrace.addScope (String)`
It is recommended not to modify core api as possible.

Issue #5392, opened 2019-03-22
Pinpoint 1.8.2 - MongoDB calls not intercepted


asgs:
1. Created a [sample application](https://github.com/asgs/pinpoint-mongodb-tests) that connects to a local MongoDB instance, queries a collection, loops through the cursor and print each document.
2. Run the application by passing the options below.
`-javaagent:/home/asgs/pinpoint-agent/pinpoint-bootstrap-1.8.2.jar -Dpinpoint.agentId=gowri-local-test1 -Dpinpoint.applicationName=gowri-local-test1`

#### What did you do to trigger the bug?
We're not able to see MongoDB component on the remote Pinpoint Web server.
#### Expected behavior
We should be able to see the MongoDB as one of the external components that the application running the agent is connecting to

#### Logs
[gowri-local-test1-pinpoint.log.zip](https://github.com/naver/pinpoint/files/2996383/gowri-local-test1-pinpoint.log.zip)

#### Additional context
Tried running the application with both mongodb-driver and mongo-java-driver (versions 3.5.0 and other versions below 3.9.0)

ychandu:
+1

RoySRose:
Hello, @asgs @ychandu 

I've created the plugin since after 3.0 has released.
Since, [their recommendation](http://mongodb.github.io/mongo-java-driver/3.9/driver/tutorials/databases-collections/) on accessing Database and Collection was through below classes
```
import com.mongodb.client.MongoCollection
import com.mongodb.client.MongoDatabase
```

I didn't instrument the ones used in 2.x
```
com.mongodb.DB
com.mongodb.DBCollection
```
So, if you are using below classes to access DB and Collections, it won't be traced.
(I'm not sure I can work on this at the time being)

Two solutions. 
You can you the APIs as Mongo Team recommended.
Or
little modification to Mongodb plugin (add `DB`,`DBCollection` to the plugin)

I'd love to guide your way if you are willing to upgrade the plugin and contribute :)

asgs:
@RoySRose  you're right, tracing works when the 3.x classes are used. Let us know how we could update the plugin to trace the 2.x classes as well.

RoySRose:
@asgs 

I've looked into the library of Mongo-driver to check classes
```
com.mongodb.DB
com.mongodb.DBCollection
```

There are 2 parts that you need to instrument.
1. Connection part (getting MongoClient, DB, Collection)
2. Session part (CRUD)
[...]

PR #5429, opened 2019-03-28
Update MongoPlugin to Intercept 2.x MongoDB driver classes

asgs:
This is a change related to #5392

---
This change is [Reviewable](https://reviewable.io/reviews/naver/pinpoint/5429)


RoySRose:
Hello, @asgs 

All right. few things.

1. Please develop based on `master` code. It's should be 1.9.0-SNAPSHOT.
2. (Optional, but recommended) Please add IntegrationTest, You can look for class *MongoDBIT_3_7_x_IT* as an example
3. Were there any interceptor code modified or added?

I'll get back to you when `1.` task is taken care of :)

asgs:
>     1. Please develop based on `master` code. It's should be 1.9.0-SNAPSHOT.

@RoySRose 

1. in 1.9.0-SS, the method `transform(String className, TransformCallback transformCallback)` is deprecated, so the code will need to be adjusted (slightly)
2. Also, since 1.9.0-SS is nightly (and might be unstable at the moment), we were inclined towards merging this change to the 1.8.x mainline and get the updated artifact relatively immediately

RoySRose:
@asgs 

I think you can use whatever you have for your current operating system.
Merging into 1.8.x mainline has no meaning since there won't be any more updates with 1.8.x.(may occur if there are critical bugs, but not likely)
Codes merged in the master will only be appied in the further versions. So, normally contributions need to be merged to master branch.

About the deprecation of method `transform`. It won't be so hard to change if you check the codes in Mongo plugin code in *master* branch.

Let me know your thoughts.

asgs:
@RoySRose makes sense.

RoySRose:
@asgs 

>shall I close this and raise a new PR based off of master?

Yeah, would be great~!

PR #5435, opened 2019-03-29
Add support to MongoDB plugin to trace the 2.x version of MongoDB Driver Classes

asgs:
Fix for https://github.com/naver/pinpoint/issues/5392

---
This change is [Reviewable](https://reviewable.io/reviews/naver/pinpoint/5435)


RoySRose:
@asgs 

I couldn't review the code, yet.
Regardless, can you add Integration Test too?

You can look for `MongoDBIT_3_7_x_IT.java` as an example.

asgs:
@RoySRose I've tried to replicate `MongoDBIT_3_7_x_IT` for 2_X as well and have hit a few test failures. Especially, an error that says [Cannot find apiId of](https://gist.github.com/asgs/3680f8bc8c2f38bde657fa220633d086)

asgs:
@RoySRose please check above in case you missed it

RoySRose:
@asgs

I'll get back to you next week :)
currently, on something urgent

RoySRose:
Hello, @asgs 

I got back. I've reviewed the code. I think you did a great job.
Few things I'd like to bring up. Tell me if I'm wrong.

1. While instrumenting to get DB `com.mongodb.Mongo.getDB()` and Collection with `com.mongodb.DB.getCollection()`. How about information on *writeConcern* and *readPreference*?

2. .
    2-1. inside `addSessionTransformer2_X`. Method retrieved with `getMethodlistR2_x` and `getMethodlistCUD2_x`. But when I've looked into **mongo-java-driver 2.13.0** there wasn't any methods named `findOneAndReplace` or `insertOne` and... more... please check the name of methods that are instrumented and clean `addSessionTransformer2_X`,  `getMethodlistR2_x` up.
    2-2. let's keep the order of the methods in one criteria. `getMethodlistR2_x` followed by `getMethodlistCUD2_x` followed by `getMethodlistR3_0_x` followed by `getMethodlistCUD3_0_x` and so on... 
3. Minor things.... let's change the name of methods for easier understand. `addMongoConnectionTransformer_2Plus` -> `addTransformer_2_0_x`
Class name `MongoConnectionTransformer_2Plus` -> `Transformer_2_0_X`
and also put comments on top of code such as. just for follwing developers. (After all it's open source project:) )
```java
//add DBConnection for 2.X and ClientConnection for 3.X 
transformTemplate.transform("com.mongodb.Mongo", Transformer_2_0_X.class);
```

Here is what I've found so far. I'll let you know if there are more.

RoySRose:
I've update the last comment.

asgs:
@RoySRose thank you! I'll try to incorporate the review comments. meanwhile, can you also check the IT error I posted above?

RoySRose:
@asgs

For, `MongoDBIT_2_X_IT`
You should reset the dependency to support all mongo-driver 2.x.
"org.mongodb:mongodb-driver:[2.0.0,2.max]"

Also for MongoDB plugin. You should check if it suuports all 2.x version.

about error
```
Cannot find apiId of [com.mongodb.DBCollection.insert(com.mongodb.DBObject[])]
```
I think this is because. 

you didn't instrument the *insert* method from the plugin.
Actually I've already mentioned above. please check `MongoPlugin.java`
```java
 private static String[] getMethodlistCUD2_x() {
        return new String[] { "dropIndexes", "dropIndex", "createIndexes", "createIndex", "updateMany", "updateOne",
                "replaceOne", "deleteMany", "deleteOne", "insertMany", "insertOne", "bulkWrite", "save", "update",
                "updateMulti" };
    }
```

asgs:
@RoySRose I've made a few changes you mentioned above and the "cannot find apiId of" error is resolved. But I'm facing [another error related to MongoDB 2.x](https://gist.github.com/asgs/8da93b17ec554b8f26c9549ff509e8f0) driver which I believe is a blocker.

1. The WriteContext class is dependent on the BsonWriter class which is available from 3.x onwards. So, I guess there needs to be another class which does this for the 2.x equivalent. But I believe it'll be a huge change and I'm not sure if I have the time and bandwidth to achieve it.

2. Even if we manage to do the above by using the equivalent class (which is the 2.x equivalent), it might NOT be available in the later versions like 3.x. So having both the classes on the Java classpath is not available. Do you have any suggestion how we could tackle this?

RoySRose:
@asgs

for both 1,2.

I think it would be best to separate `MongoRSessionInterceptor` and `MongoCUDSessionInterceptor` classes.

which makes it possible for *MongoUtil.parseBson* to be rewritten.
creating a new logic for parsing Bson. (Since `org.bson.conversions.Bson` doesn't exist in 2.x anyway)
```java
if (args != null) {
         NormalizedBson parsedBson = MongoUtil.parseBson(args, traceBsonBindValue);
         MongoUtil.recordParsedBson(recorder, parsedBson);
 }
```

I don't think a **new logic for parsing Bson** will be difficult. Since it can be based on the original  need a slight change *MongoUtil.recordParsedBson* method with slight changes.

In this way. Both 1,2 problems will be solved.

asgs:
@RoySRose thank you for your suggestions. I made the changes as asked and it worked for most of the methods.

But I'm getting an error [Expected a SpanEvent with id x but was y ](https://gist.github.com/asgs/6f27778bbe7bdb79eb35b503beb7fecc) happens while tracing the methods `find(DBObject)` and `remove(DBObject)`. it dd work for the no-arg `find()` method, so I'm not sure what is going wrong here

asgs:
@RoySRose Ignore the error I posted above. It was caused by tracing a few methods that are abstract. I've updated the tests and pushed all the commits.

Please review the code. Hoping to have this PR merged at the earliest!

asgs:
@RoySRose I hope you're looking into this PR. let me know if you have any issues, please

RoySRose:
Hello, @asgs 

I've managed to make some time for the review.
I think you've understood the concept of the plugin. I'll get into the details.

First, I'll go through `MongoPlug.java` and continue to review on the interceptors.

#### MongoPlug.java
 1. I think you should leave methods that concern 3.x~ as it is. Since `ClientConnectionTransform3_0_X` method will be disabled in this PR. So the best structure would be 

```java
        addConnectionTransformer2_X();
        addConnectionTransformer3_0_X();
        addConnectionTransformer3_7_X();
        addConnectionTransformer3_8_X();

        addSessionTransformer2_X();
        addSessionTransformer3_0_X();
        addSessionTransformer3_7_X();

```
and leave `addConnectionTransformer3_0_X` method as it is. And `addMongoConnectionTransformer_2_X` method should go inside `addConnectionTransformer2_X` method, such as
```java
private void addConnectionTransformer2_X() {
        transformTemplate.transform("com.mongodb.Mongo", MongoConnectionTransformer_2_X.class);
        transformTemplate.transform("com.mongodb.DB", ConnectionTransformer2_X.class);
    }
```

 2. Since 2.x doesn't need client connection like 3.x
changed interceptor names like 3.x will be better to understand 
MongoConnectionTransformer_2_X -> DatabaseConnectionTransform2_X
ConnectionTransformer2_X -> CollectionConnectionTransform2_X

 3. You don't need to try-catch `NotFoundInstrumentException` after you separate 2.x and 3.x connection (**1.** that I've mentioned above). You can just let it throw the exception to handle from the upper class.

 4. `MongoDriverConnectInterceptor2_X` should not be called for method readPreference, setWriteConcern method. For this, I think you can check if `MongoReadPreferenceInterceptor` and `MongoWriteConcernInterceptor` can be re-used.

 5. About instrumenting the constructor of "com.mongodb.Mongo" class. You should instrument Constructor `Mongo(MongoAuthority authority, MongoOptions options)`, not `Mongo(String host, int port)` because some constructors don't bypass `Mongo(String host, int port)` But all of them bypass `Mongo(MongoAuthority authority, MongoOptions options)`

Let me know if there is anything I've misunderstood.

RoySRose:
@asgs

notifying you in case you've missed.

asgs:
@RoySRose sorry, couldn't get time to attend to this. will try to see if I can address these comments this week

asgs:
@RoySRose If we keep the `addConnectionTransformer3_0_X()` untouched, then we need to transform the `com.mongodb.Mongo` class twice, once for 2_x and again for 3_x. If I remember correctly, there was an error that the class is already transformed.

Hence, I transformed it exactly once handling both 2_x and 3_x in one shot. So, I'm not sure if your comments 1 and 3 can be addressed. Let me know if I'm missing something

RoySRose:
I'll be on a business trip this week.
I'll get back to this, first thing next week.

asgs:
> I'll be on a business trip this week.
sure, have a nice one!

RoySRose:
>If I remember correctly, there was an error that the class is already transformed.

RoySRose:
Hello, @asgs 
I'm back

>If we keep the addConnectionTransformer3_0_X() untouched, then we need to transform the com.mongodb.Mongo class twice, once for 2_x and again for 3_x. If I remember correctly, there was an error that the class is already transformed.

Yes, You're right. There will be a problem when trying to transform classes twice. I'll look for a way to distinguish the versions and separate from the instrumenting stage.

>Hence, I transformed it exactly once handling both 2_x and 3_x in one shot. So, I'm not sure if your comments 1 and 3 can be addressed. Let me know if I'm missing something

3.x is not working with these changes. So it looks like it needs some fix.

>Regarding comment 4, I used to get some error regarding reflection when I tried to reuse those interceptors, but I don't remember it. Let me reproduce the error and update this comment.

Please do check

>Regarding comment 5, the class MongoAuthority was introduced in v2.11.0 and it's not a public class for some reason. I'm not sure if we can depend on this class unless we are fine with reducing the range of MongoDB driver versions we want to support on the Integration tests

I don't fully understand why. Even it was introduced since v2.11.0. It should be taken care of for v2.11.0 and after.
I'm looking in the library of v2.14.0 and all constructors of `com.mongodb.Mongo` class eventually calls `Mongo(MongoAuthority authority, MongoOptions options) {...`  so this should be used for instrument. If it's different for versions before v2.11.0 than it also should be taken care of.

RoySRose:
Using the above reflection may not work due to class loading time.
If this is the case try out below.
Below code should be inside the transformer.
(If below is used it would be good to change the name of `ClientConnectionTransform3_0_X` method to `ClientConnectionTransform` or anything that fits better.

```java
      target.addField(DatabaseInfoAccessor.class);

            InstrumentMethod connectVersion2 = target.getConstructor("java.lang.String","int");
            if (connectVersion2 != null){
                connectVersion2.addScopedInterceptor(MongoDriverConnectInterceptor2_X.class, MONGO_SCOPE,
                        ExecutionPolicy.BOUNDARY);
            }

            InstrumentMethod connectVersion3 = target.getConstructor("com.mongodb.connection.Cluster", "com.mongodb.MongoClientOptions", "java.util.List");
            if (connectVersion3 != null){
                connectVersion3.addScopedInterceptor(MongoDriverConnectInterceptor2_X.class, MONGO_SCOPE,
                        ExecutionPolicy.BOUNDARY);
            }
```

asgs:
@RoySRose I have fixed comment 4. Thankfully, I didn't receive any reflection-related errors this time, so I pushed the commits. I've fixed comment 1 partially which is regarding the naming conventions and the grouping of the methods.

Let me take a look at your suggestions for the rest of the comments.

asgs:
@RoySRose As you mentioned, checking the `MAJOR_VERSION` doesn't work during `setup()` as the MongoDB driver won't be loaded yet. But the other workaround you mentioned regarding the driver connect interceptor worked perfectly. The small thing is, we just need to add either of the interceptors and try to add the 3.x specific constructor first, followed by the 2.x's since the 2.x constructor continues to be present in the 3.x drivers

This way, whatever range of MongoDB driver versions we're intending to support on the ITs will continue to be supported. Looks like a win-win. Can you please review?

RoySRose:
Great job! with good ideas.

I'll point out few things first :)

1. I don't think `com.mongodb.DBCollection` need to be transformed since it's abstract class and will always be approached from `com.mongodb.DBCollectionImpl` anyway.

```java
private void addSessionTransformer2_X() {
        // java driver 2.X
        transformTemplate.transform("com.mongodb.DBCollection", SessionTransformer2_X.class);
        transformTemplate.transform("com.mongodb.DBCollectionImpl", SessionTransformer2_X.class);
    }
```

2. Current `MongoPlugin.java` still instrudes v3.
For v3, methods
 - setReadPreference
 - setWriteConcern
 - getDB
doesn't need to be instrumented since they are all deprecated.
so I suggest below code to be moved inside the ifelse statement above.(inside v2 scope)

```java
    InstrumentMethod getReadPreference = InstrumentUtils.findMethod(target, "setReadPreference", "com.mongodb.ReadPreference");
            getReadPreference.addScopedInterceptor(MongoReadPreferenceInterceptor.class, MONGO_SCOPE, ExecutionPolicy.BOUNDARY);

            InstrumentMethod getWriteConcern = InstrumentUtils.findMethod(target, "setWriteConcern", "com.mongodb.WriteConcern");
            getWriteConcern.addScopedInterceptor(MongoWriteConcernInterceptor.class, MONGO_SCOPE, ExecutionPolicy.BOUNDARY);

         
//... and....

            try {
                InstrumentMethod database = InstrumentUtils.findMethod(target, "getDB", "java.lang.String");
                database.addScopedInterceptor(MongoDriverGetDatabaseInterceptor.class, MONGO_SCOPE,
                        ExecutionPolicy.BOUNDARY);
            } catch (NotFoundInstrumentException e) {
                if (LOGGER.isDebugEnabled()) {
                    LOGGER.debug("2.x method getDB not found. Ignoring...");
                }
            }
```

3. Inside `MongoRSessionInterceptor2_X.java`. 
StringBuilder builder = new StringBuilder(); is not used anywhere. does it?

4. VERY IMPORTANT THING! (I've looked into library v2.14.2)
For v2. constructor  `Mongo(MongoAuthority authority, MongoOptions options)` should be instrumented in `com.mongodb.Mongo` not the current one

```java
InstrumentMethod connectVersion2 = target.getConstructor("java.lang.String", "int");
-->
InstrumentMethod connectVersion2 = target.getConstructor("xx.xx.MongoAuthority", "xx.xx.MongoOptions");
```
For example, with current one, when Mongo is constructed with `List<ServerAddress> seeds`. Tracing will not work

5. I'll check `parseJson` method you've created in `MongoUtil` class.
This will take some time since I need to dig in deeper. But already I can see few modifications are needed.
For example `traceJsonBindValue` should be used.
But I'll get back when I've fully reviewed on the method.

RoySRose:
I'd like to ask you one thing. which version of the library are you using?

Because even v2.14 is very similar to v3.x
https://mongodb.github.io/mongo-java-driver/2.14/getting-started/quick-tour/

asgs:
@RoySRose 
> I don't think com.mongodb.DBCollection need to be transformed since it's abstract class and will always be approached from com.mongodb.DBCollectionImpl anyway.
I also thought about instrumenting the `DBCollectionImpl` class alone, but I was getting the `cannot find apiId of` error which means `DBCollection` also needs to be instrumented for some reason. It happens with 2.x driver related ITs. So I ended up instrumenting both

> Current MongoPlugin.java still instrudes v3.
Makes sense

> Inside MongoRSessionInterceptor2_X.java.
StringBuilder builder = new StringBuilder(); is not used anywhere. does it?
Yeah, will remove this.

>For v2. constructor Mongo(MongoAuthority authority, MongoOptions options) should be instrumented in com.mongodb.Mongo not the current one
`MongoAuthority` was introduced from 2.11 onwards. v2.7.0 till 2.10 use the constructor `Mongo(ServerAddress addr, MongoOptions options)`. so, this means we'll have to reduce the range of MongoDB driver versions supported as we won't be able to lookup the class during `setup()`

> I'll appreciate if you can show me a screenshot that you've tested with v2.
I've not tested live APM since last few months. will do it some time later

RoySRose:
>MongoAuthority was introduced from 2.11 onwards. v2.7.0 till 2.10 use the constructor Mongo(ServerAddress addr, MongoOptions options). so, this means we'll have to reduce the range of MongoDB driver versions supported as we won't be able to lookup the class during setup()

Got it. I've just downloaded v.2.7.3 and looking into it.
As you have said for v2.7.0 till 2.10 below constructor should be instrumented.
Mongo(ServerAddress addr, MongoOptions options)

And here is more trick part. NOW, you have to distinguish v2.10 ~ v2.12  too. Unless v2.10 ~ v2.12 is taken care of with the plugin code for v3.x

RoySRose:
Due to the difference of v2.7.0 ~ v2.10 and v2.10 ~ v2.12. The plugin may become very complex.
I think there will be a way if we take a close look at it. 
But, If not possible. separating the plugin itself can also be one of the solutions.
(ex. `Pinpoint-mongodb-driver-plugin`, and a new `Pinpoint-mongodbv2-driver-plugin`)

But I still recommend the first one. Because the libraries of mongo java driver v2.x are fixed and not going to be updated in the future.

RoySRose:
@asgs 
any updates?

asgs:
> any updates?
Sorry, I wasn't able to focus on this. But a quick question - how would you lookup the MongoDB Driver library version at runtime? I think we discussed this topic a few weeks back and concluded it's not possible to do so.

At the point, I feel like this PR has lived for a little more longer than I thought. Could we restrict the versions supported to v2.11-higher for now and open another PR to see the possibilities of extending the range of versions?

RoySRose:
Hello, @asgs 

>@RoySRose Sorry, I wasn't able to focus on this. 

No Problem :) nobody is rushing, it's open-source!

> But a quick question - how would you lookup the MongoDB Driver library version at runtime? I think we discussed this topic a few weeks back and concluded it's not possible to do so.

Well, It's not impossible to look up the version of MongoDB Driver library at runtime.
But, It needs some modification in the core logic of Pinpoint and it's not ready for it.

> At this point, I feel like this PR has lived for a little more longer than I thought. Could we restrict the versions supported to v2.11-higher for now and open another PR to see the possibilities of extending the range of versions?

Restricting version support for v2.11 and higher is perfectly fine with me~ :)

asgs:
> Well, It's not impossible to look up the version of MongoDB Driver library at runtime.
> But, It needs some modification in the core logic of Pinpoint and it's not ready for it.

Yeah, that's what I tried to mean, too.

> Restricting version support for v2.11 and higher is perfectly fine with me~ :)

Perfect! I'll update the code to instrument the correct Mongo constructor and also update the version range.

RoySRose:
@asgs Let me know if your update is finished.

I'll take a look and run some tests

asgs:
> @asgs Let me know if your update is finished.

@RoySRose I didn't try anything after the non-public class `MongoAuthority` couldn't be imported. Looks like I'm stuck

asgs:
> I'll take a look and run some tests

@RoySRose just checking if you had any updates

RoySRose:
@asgs 

I had my baby come out 3weeks ago and went on a business trip in taiwan, came back this morning.

I'll check on this asap :)

plugins/mongodb/src/main/java/com/navercorp/pinpoint/plugin/mongo/MongoUtil.java
```diff
@@ -86,5 +95,66 @@ public static NormalizedBson parseBson(Object[] args, boolean traceBsonBindValue
...
+                DBObject[] objects = (DBObject[]) arg;
+                for (DBObject object : objects) {
+                    extractFromDbObject(parsedJsonBuilder, jsonParamBuilder, object);
+                }
+                adjustJson(parsedJsonBuilder, jsonParamBuilder);
+            } else if (arg instanceof DBObject) {
+                extractFromDbObject(parsedJsonBuilder, jsonParamBuilder, (DBObject) arg);
+                adjustJson(parsedJsonBuilder, jsonParamBuilder);
+            } else {
+                System.err.println("Not a DBObject; it's a(n) " + arg.getClass());
```
RoySRose: System.err ~~ should be taken care of with something else such as log

plugins/mongodb/src/main/java/com/navercorp/pinpoint/plugin/mongo/MongoUtil.java
```diff
@@ -86,5 +95,66 @@ public static NormalizedBson parseBson(Object[] args, boolean traceBsonBindValue
...
+            }
+        }
+
+        String parsedJsonString = parsedJsonBuilder.toString();
+        String jsonParameterString = jsonParamBuilder.toString();
+        System.out.println("parsedJsonString - " + parsedJsonString + "; jsonParameterString - " + jsonParameterString);
```
RoySRose: same here
System.err ~~ should be taken care of with something else such as log

plugins/mongodb/src/main/java/com/navercorp/pinpoint/plugin/mongo/MongoUtil.java
```diff
@@ -86,5 +95,66 @@ public static NormalizedBson parseBson(Object[] args, boolean traceBsonBindValue
...
+                continue;
+            }
+            parsedJsonBuilder.append("\"");
+            jsonParamBuilder.append("\"");
+            parsedJsonBuilder.append(key);
+            parsedJsonBuilder.append("\"");
+            parsedJsonBuilder.append(":");
+            parsedJsonBuilder.append("\"");
+            parsedJsonBuilder.append("?");
+            jsonParamBuilder.append(object.get(key));
```
RoySRose, line 148: collecting value should depend on `traceJsonBindValue`

plugins/mongodb/src/main/java/com/navercorp/pinpoint/plugin/mongo/MongoUtil.java
```diff
@@ -86,5 +95,66 @@ public static NormalizedBson parseBson(Object[] args, boolean traceBsonBindValue
...
+            parsedJsonBuilder.append("\"");
+            jsonParamBuilder.append("\"");
+            parsedJsonBuilder.append(",");
```
RoySRose: if this ends with `,` 

doesn't the String be something like "a,b,c," ?

Please check once again

RoySRose: I can see the last "," is taken care of from adjustJson method.
But I think this is an unnecessary procedure. What do you think?
I think it would be better not to add last "," from the first place.

plugins/mongodb/src/main/java/com/navercorp/pinpoint/plugin/mongo/MongoUtil.java
RoySRose: Also, won't object.get(key) will again iterate all key data in object?
For the plugins. We are sensitive in performance

checkout StringJoiner.class for few hints

RoySRose:
@asgs
Let me know what you think about the review

asgs:
> I had my baby come out 3weeks ago and went on a business trip in taiwan, came back this morning.

Many Congratulations and hearty wishes to you and your family!

>  Let me know what you think about the review

sure, I'll check and revert

asgs:
@RoySRose fixed the comments shared earlier and resolved the conversation relating to those.

We're still undecided on how to handle this part because `MongoAuthority` is not a public class we could import
> For v2. constructor Mongo(MongoAuthority authority, MongoOptions options) should be instrumented in com.mongodb.Mongo not the current one

RoySRose:
```java
public static NormalizedBson parseJson(Object[] args, boolean traceJsonBindValue) {
```

*traceJsonBindValue* need to be used.

This option decides collect binded value or not

RoySRose:
>We're still undecided on how to handle this part because MongoAuthority is not a public class we could import
For v2. constructor Mongo(MongoAuthority authority, MongoOptions options) should be instrumented in com.mongodb.Mongo not the current one

Checking [official documents](https://mongodb.github.io/mongo-java-driver/2.13/getting-started/quick-tour/)

shouldn't MongoClient be instrumented? instead of Mongo class?
image


All the API to Mongo(MongoAuthority authority, MongoOptions options) is deprecated.
Since we are just supporting since 2.13 and 2.14, does Mongo class need to be instrumented?

Let me hear what you think.

RoySRose:
@asgs 

Let me know if you have any updates

asgs:
@RoySRose sorry, couldn't catch up here. Will do this by next week

asgs:
>  Since we are just supporting since 2.13 and 2.14, does Mongo class need to be instrumented?

@RoySRose we were originally planning to support v2.7 and we encountered few issues and then we bumped it to v2.9

So, if we're only looking to support v2.13 and the rest few, I'll do the changes accordingly. It'll require some changes to the `MongoPlugin` class and the corresponding driver connect interceptor. Might have to also refactor the IT base class.

Let me know if we're good to proceed

RoySRose:
Hello, @asgs 

Good to hear from you.
Your will is all that need (since it's open source :) )
If you wish to only support v2.13 and higher in your contribution, it's fine with me.

Hope to see your contribution. 
Thank you.

asgs:
@RoySRose Long time! I just pushed a commit instrumenting the MongoClient class instead of Mongo class as discussed last time.

Please review and let me know your thoughts and comments. Sorry about the delay. I'm unable to focus on this PR on a constant basis

asgs:
@RoySRose PING (just in case you missed the comment above)

RoySRose:
@asgs 

Oh, I do have this in mind. Just a little busy on a few things.
And... yes, it's been a while. So I need to reload to my head, what was going on in this PR.

Hope to take care of this next week :)

RoySRose:
Hello, @asgs 

Sorry, for such a late reply.

I've managed to make time to take a look at your code.
Eventually, the result may somehow be the same, but this holds a high potential for future problems.

For example. `ClientConnectionTransform2_X_3_0_X` class.
First half which is for v3 (and up) looks fine.
But below code have some risks.
```java
try {
                InstrumentMethod database = InstrumentUtils.findMethod(target, "getDB", "java.lang.String");
                database.addScopedInterceptor(MongoDriverGetDatabaseInterceptor.class, MONGO_SCOPE,
                        ExecutionPolicy.BOUNDARY);
            } catch (NotFoundInstrumentException e) {
                LOGGER.info("Couldn't find the 2.x's getDB(String) method from the class " + className);
            }
```

for users using v3 and up. `getDB()` should not be instrumented. But from this code. It will be instrumented(which is a deprecated function in v3 and up)

Here is my suggestion.

1. Since this PR is quite old. Let create a new PR for refreshing. And while doing that It would be better to squash relevant commit into fewer(3,4?) commits, rather than 24 commit as now.
Also, commits needs to be `rebase` instead of `merge` as of this opensource project policy. If you don't find the difference, this article may help you. (https://medium.com/datadriveninvestor/git-rebase-vs-merge-cc5199edd77c)
Below is one of the pictures in the article 

image

2. Since your changes(commits) still have affects on v3(and up). I suggest you test v3 also after changing the code. IF you find this hard due to the circumstances, I think we can postpone this PR for a few months later since we are preparing to separate loading plugins by their versions(for example, for the application using v2 mongo only loads pluginv2 and for the application using v3 mongo only load pluginv3). After this feature is developed, we don't have to worry about any collisions just like we are facing right now.

Let me hear your thoughts :) have a nice day~

asgs:
@RoySRose Ahoy!

> getDB() should not be instrumented.

Any reason why? If it shouldn't be, I guess we could return the bytecode before executing this block of code, unless `connectVersion3` is null

>  we are preparing to separate loading plugins by their versions

I guess we should wait for this to happen. Don't want to spend another cycle reworking the PR at this time

RoySRose:
>Any reason why? If the method doesn't exist in future versions (v3+), it would be caught and logged. If it still shouldn't be instrumented, I guess we could return the bytecode before executing this block of code, unless connectVersion3 is null

Well, for v3, the method is deprecated and the plugin is already designed and fully functional without using deprecated functions. So there is no need to instrument `getDB()` 

>I guess we should wait for this to happen. Don't want to spend another cycle reworking the PR at this time

Understood. No need to rush things up if there is no need. 

I'll close the PR :)

asgs:
> for v3, the method is deprecated and the plugin is already designed and fully functional without using deprecated functions. So there is no need to instrument getDB()

Yeah, but instrumenting this method is required for V2. So, this code will be instrumented conditionally. Hope that is fine.

RoySRose:
Yes. Instrumenting it conditionally would be okay. But currently for some part of the code it's always instrumenting.

asgs:
> Instrumenting it conditionally would be okay

Perfect! Thank you

yjqg6666:
@asgs @RoySRose This PR would not been merged? Any update?

RoySRose:
Unfortunately no, as you can't see. It not merged.

If issue #7078 clears out. It will be much easier to take care of this kind of PR

Issue #5482, opened 2019-04-11
Add Spring WebFlux Plugin


jaehong-kim:
Add Spring WebFlux Plugin
* org.springframework:spring-webflux 5.x
* Server & Client

~~~
###########################################################
# Spring WebFlux
###########################################################
profiler.spring.webflux.enable=true

# Client
# Record Parameter.
profiler.spring.webflux.client.param=true
# Record cookies.
profiler.spring.webflux.client.cookie=true
# When cookies should be dumped. It could be ALWAYS or EXCEPTION.
profiler.spring.webflux.client.cookie.dumptype=ALWAYS
# 1 out of n cookies will be sampled where n is the rate. (1: 100%)
profiler.spring.webflux.client.cookie.sampling.rate=1
# Cookie dump size.
profiler.spring.webflux.client.cookie.dumpsize=1024
~~~

PR #5483, opened 2019-04-11
[#5482] Add Spring WebFlux Plugin

jaehong-kim:
#5482


Roronoa-Zoro:
@jaehong-kim  hi Kim, when will pinpoint team release 1.9.0 version?

PR #5506, opened 2019-04-18
Alias feature

RoySRose:
To take care of servermap displayed when embedded library exists.
Such as HTTP4 library embedded in Elasticsearch client library

Only for RPC libraries. (Feature unnecessary for others)
Following terminal node may or may not exist (has to cover both cases)

### DIFF
--- a/collector/src/main/java/com/navercorp/pinpoint/collector/service/TraceService.java
+++ b/collector/src/main/java/com/navercorp/pinpoint/collector/service/TraceService.java
@@ -22,8 +22,8 @@ import com.navercorp.pinpoint.collector.dao.TraceDao;
 import com.navercorp.pinpoint.common.server.bo.SpanBo;
 import com.navercorp.pinpoint.common.server.bo.SpanChunkBo;
 import com.navercorp.pinpoint.common.server.bo.SpanEventBo;
-import com.navercorp.pinpoint.loader.service.ServiceTypeRegistryService;
 import com.navercorp.pinpoint.common.trace.ServiceType;
+import com.navercorp.pinpoint.loader.service.ServiceTypeRegistryService;
 import org.apache.commons.collections.CollectionUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -31,6 +31,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
 import java.util.List;
+import java.util.Objects;
 
 @Service
 public class TraceService {
@@ -72,6 +73,19 @@ public class TraceService {
         insertAcceptorHost(spanBo);
         insertSpanStat(spanBo);
         insertSpanEventStat(spanBo);
+    }
+
+    private void insertAcceptorHost(SpanEventBo spanEvent, String applicationId, ServiceType serviceType) {
+
+        String endPoint = spanEvent.getEndPoint();
+        if(endPoint == null){
+            return;
+        }
+        String destinationId = spanEvent.getDestinationId();
+        if(destinationId == null){
+            return;
+        }
+        hostApplicationMapDao.insert(endPoint, destinationId, spanEvent.getServiceType(), applicationId, serviceType.getCode());
     }
 
     private void insertAcceptorHost(SpanBo span) {
@@ -177,11 +191,20 @@ public class TraceService {
     }
 
     private void insertSpanEventList(List<SpanEventBo> spanEventList, ServiceType applicationServiceType, String applicationId, String agentId, String endPoint) {
+
         for (SpanEventBo spanEvent : spanEventList) {
             final ServiceType spanEventType = registry.findServiceType(spanEvent.getServiceType());
+
+            if (spanEventType.isAlias()) {
+                insertAcceptorHost(spanEvent, applicationId, applicationServiceType);
+            }
+
             if (!spanEventType.isRecordStatistics()) {
                 continue;
             }
+
+            final String spanEventApplicationName = spanEvent.getDestinationId();
+            final String spanEventEndPoint = spanEvent.getEndPoint();
 
             // if terminal update statistics
             final int elapsed = spanEvent.getEndElapsed();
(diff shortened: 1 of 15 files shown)


collector/src/main/java/com/navercorp/pinpoint/collector/service/TraceService.java
```diff
@@ -74,6 +74,10 @@ public void insertSpan(final SpanBo spanBo) {
         insertSpanEventStat(spanBo);
     }
 
+    private void insertAcceptorHost(SpanEventBo spanEvent, String applicationId, ServiceType serviceType) {
+        hostApplicationMapDao.insert(spanEvent.getEndPoint(), spanEvent.getDestinationId(), spanEvent.getServiceType(), applicationId, serviceType.getCode());
```
Xylus: Null check here for end point and destination id might be a good idea.

web/src/main/java/com/navercorp/pinpoint/web/service/map/processor/RpcCallProcessor.java
```diff
@@ -109,6 +110,29 @@ public LinkDataMap processLinkDataMap(LinkDataMap linkDataMap, Range range) {
...
+
+        for (AcceptApplication acceptApplication : acceptApplicationList) {
+            if (!acceptApplication.getApplication().getServiceType().isAlias()) {
+                resultSet.add(acceptApplication);
+            } else {
+                logger.debug("deduct alias application {}", acceptApplication);
+            }
+        }
+
+        if (resultSet.size() == 0) {
```
Xylus: Would `isEmpty()` suffice?

web/src/main/java/com/navercorp/pinpoint/web/service/map/processor/RpcCallProcessor.java
```diff
@@ -130,11 +154,15 @@ public LinkDataMap processLinkDataMap(LinkDataMap linkDataMap, Range range) {
         AcceptApplicationCacheKey cacheKey = new AcceptApplicationCacheKey(fromApplication, range);
         Set<AcceptApplication> cachedAcceptApplications = acceptApplicationCache.get(cacheKey);
         if (cachedAcceptApplications == null) {
-            logger.debug("acceptApplicationCache hit {}", fromApplication);
+            logger.debug("acceptApplicationCache miss {}", fromApplication);
```
Xylus, line 157: Good catch :+1:

collector/src/main/java/com/navercorp/pinpoint/collector/service/TraceService.java
```diff
@@ -177,12 +181,21 @@ private void insertSpanEventStat(SpanBo span) {
     }
 
     private void insertSpanEventList(List<SpanEventBo> spanEventList, ServiceType applicationServiceType, String applicationId, String agentId, String endPoint) {
+
         for (SpanEventBo spanEvent : spanEventList) {
             final ServiceType spanEventType = registry.findServiceType(spanEvent.getServiceType());
+
+            if (spanEventType.isAlias()) {
```
Xylus, line 201: If aliased **ServiceType** needs to be `recordStatistics=true` as well, you might want to place this check after `if (!spanEventType.isRecordStatistics())`.
If it doesn't, please ignore this comment.

web/src/main/java/com/navercorp/pinpoint/web/service/map/processor/RpcCallProcessor.java
RoySRose: I don't understand what you're suggesting.
`filterAlias` method is trying to remove aliased application types if there are application types from non-alias serviceTypes. 
This logic is used when there are both client-plugin and server-plugin

collector/src/main/java/com/navercorp/pinpoint/collector/service/TraceService.java
RoySRose: aliased ServiceType need to be`recordStatistics=false`

collector/src/main/java/com/navercorp/pinpoint/collector/service/TraceService.java
RoySRose: Is it necessary? since there are going to be checked inside `HbaseHostApplicationMapDao.insert()` method.

web/src/main/java/com/navercorp/pinpoint/web/service/map/processor/RpcCallProcessor.java
Xylus: Nothing so extensive, was suggesting for clarity, `if (resultSet.size() == 0) {` -> `if (resultSet.isEmpty()) {`

RoySRose: oh, got it

collector/src/main/java/com/navercorp/pinpoint/collector/service/TraceService.java
Xylus: The check inside `HbaseHostApplicationMapDao.insert()` throws a **NullPointerException**, so it will skip all span event statistics that come after this from being stored.

RoySRose: taken care of

RoySRose:
@Xylus added tests and restraints.

collector/src/main/java/com/navercorp/pinpoint/collector/service/TraceService.java
```diff
@@ -177,12 +191,21 @@ private void insertSpanEventStat(SpanBo span) {
...
+                insertAcceptorHost(spanEvent, applicationId, applicationServiceType);
```
Xylus, line 202: I think it would be safer to add a `continue;` here.
Even if the `if (!spanEventType.isRecordStatistics())` check is right below, but I think adding an explicit `continue;` here would safeguard against malformed cases where service type has both `AS_ALIAS` and `RECORD_STATISTICS`.
What do you think?

RoySRose: There is a reason why I didn't add `continue;` there.

Since restraints are set in `TraceMetadataLoader.java`.
Logically, `continue;` doesn't have any effects.

Rather, it may be more reasonable to think that it should be included because it doesn't matter.

But I thought it would be a good idea to put the constraints together, not scattered across the codes. So If the day comes to remove those constraints. we can just take a look at the loading part.

What do you think?

Xylus: Sounds fair enough~

RoySRose:
@emeroad
anything else?

PR #5551, opened 2019-05-02
[#5411] Add histogram for Alias

RoySRose:
Add histogram data for As_Alias option.
This PR is to fix https://github.com/naver/pinpoint/pull/5546#issuecomment-488020654

@Xylus please review when you have time. Love to hear if there is any potential problem that I'm not ware of.


yin-bp:
@RoySRose 
I have try out this fix,It's very perfect to fix bug in #5546 (comment) 

new PR is #5582

RoySRose:
Removed code in LinkSerializer to set the flow same with `Terminal` nodes.

PR #5731, opened 2019-06-24
support apache-dubbo-2.7.0

qq295190549:
support apache-dubbo-2.7.0,and processing dubbo rpcContext.getLocalAddressString() to get the local host problem in a multi-NIC environment.


qq295190549:
support apache-dubbo-2.7.0

PR #5834, opened 2019-08-05
Tomcat JDBC support

kwangil-ha:


plugins/tomcat-jdbc/src/main/java/com/navercorp/pinpoint/plugin/tomcat/jdbc/interceptor/DataSourceConstructorInterceptor.java
```diff
@@ -0,0 +1,63 @@
...
+    public void before(Object target, Object[] args) {
+    }
+
+    @Override
+    public void after(Object target, Object[] args, Object result, Throwable throwable) {
+        if (!InterceptorUtils.isSuccess(throwable)) {
+            return;
+        }
+
+        final DataSource dataSource = getDatasource(target);
```
Xylus, line 47: Referencing **DataSource** here directly may cause issues in future versions if the class name or namespace changes.
This however does seem very unlikely to happen so it should be okay here. Just dropping a line for future reference.

PR #5965, opened 2019-09-04
fix sql comment, mysql syntax error

yjqg6666:
When creating tables from the sql script, syntax errors are found. Fixed in this PR.


minwoo-jung:
good!

PR #5984, opened 2019-09-11
integration support with external log system for pinpoint web 

yjqg6666:
* expose applicationName & spanId to log.page.url
* add config log.transaction.info.always.logged config to enable log button regardless of the data.
* add config log.page.new.window to open the log.page.url in a new window.

Info:
The full URL of the log page would be ${log.page.url}?transactionId=XXX&spanId=xxx&applicationName=xxx&time=xxx. If your external log system, i.e. ELK use different query parameter, you could rewrite the url via nginx location.

> log.page.url=XXXX.pinpoint

```
	location = /XXXX.pinpoint {
		return 301 https://elk.example.com/query?app=$arg_applicationname&txId=$arg_transactionid&spanid=$arg_spanid&ts=$arg_time;
	}
```

### DIFF
--- a/doc/per-request_feature_guide.md
+++ b/doc/per-request_feature_guide.md
@@ -247,6 +247,22 @@ log.enable= true
 log.page.url=XXXX.pinpoint
 log.button.name= log
 ```
+The log page url would be /XXXX.pinpoint?transactionId=XXX&spanId=xxx&applicationName=xxx&time=xxx .
+If you want to integrate with external log system, you may add the following configuration and configure Nginx to relocate to the system url.
+```properties
+# enable log button regardless of the data.
+log.transaction.info.always.logged = true
+# open the log.page.url in a new page.
+log.page.new.window = true
+```
+
+```
+	location = /XXXX.pinpoint {
+	    # change the url as necessary
+		return 301 https://elk.example.com/query?app=$arg_applicationname&txId=$arg_transactionid&spanid=$arg_spanid&ts=$arg_time;
+	}
+```
+
 
 **step 3**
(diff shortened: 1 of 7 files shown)


minwoo-jung:
### 1

I recommend to configure redirects on the pinpoint-web server.
you should implement controller to redirect external site.

ref [guide](https://github.com/naver/pinpoint/blob/master/doc/per-request_feature_guide.md#3-expose-log-in-pinpoint-web)
example )
```
@Controller
public class ELKLogController {
  

  
    @RequestMapping(value = "/elkLog", method = RequestMethod.GET)
    public RedirectView redirectELKsite(@RequestParam(value= "transactionId", required=true) String transactionId,
                                            @RequestParam(value= "spanId", required=false) String spanId,
                                            @RequestParam(value="time", required=true) long time) {
        StringBuilder urlBuilder = new StringBuilder();
        urlBuilder.append(ELKSightUrl);
        urlBuilder.append(makeParameter(transactionId, time));
        RedirectView view = new RedirectView(urlBuilder.toString());
        
        return view;
```

### 2

The logged option(log.transaction.info.always.logged ) in pinpoint-web should not be provided.
Whether or not logging is done must be determined in the agent. it should be not make a decision on the pinpoint web.
please ref [step3 of guide](https://github.com/naver/pinpoint/blob/master/doc/per-request_feature_guide.md#3-expose-log-in-pinpoint-web)

yjqg6666:
updated PR #5996

yjqg6666:
Now the content of the log.page.url would be shown in a popup. There is no way to configure it to be open in a new window.

PR #5997, opened 2019-09-17
Fix self-impl. log controller impossible to get RequestParam spanId and applicationName

yjqg6666:
For PinPoint 1.8.x only.


minwoo-jung:
@yjqg6666 
Unfortunately, this feature doesn't seem to be backported. The backport target only targets bug fixes or critical features.

yjqg6666:
This one could be treated as a bugfix. As the doc stated "https://github.com/naver/pinpoint/blob/master/doc/per-request_feature_guide.md#3-expose-log-in-pinpoint-web"

> @Controller

> public class Nelo2LogController {
>   
>     @RequestMapping(value = "/????")
>     public String NeloLogForTransactionId(@RequestParam (value= "transactionId", required=true) String transactionId,
>                                             @RequestParam(value= "spanId" , required=false) String spanId,
>                                             @RequestParam(value="time" , required=true) long time ) {
> 
>           // you should implement the logic to retrieve your agent’s logs.
>     }

Now the self-impl. could not get spanId.

Issue #5343, opened 2019-03-13
Ways to increase agent id character limit


Xylus:
#### Description
Agent id is currently limited to 24 characters. This makes it hard at times to set it to a value provided by certain environment (such as when deploying onto kubernetes).
A few ways to increase this limit are as follows.

1. Hard increase the current limit to 36. Requires the least amount of work, but will break compatibility.
2. Convert kubernetes pod id into long long and use this internally. K8s seems to use RFC 4122 UUID to generate pod id, which can be converted into a long long so it'll integrate cleanly. There still needs to be a way to figure out if a given agent id is an actual value, or a long long from a UUID string.
3. Map agent ids into longs and use this internally. Requires the most amount of work, and will break compatibility, but would work best out of all solutions.

Still brain storming ideas so any suggestions are welcome.

related issues - #4638 #3504

dawidmalina:
@Xylus any updates on this? I'm afraid that soon pinpoin will lose its momentum to gain even more interests. Kubernetes deployments are now industry standard (including corporate) and pinpoint shall take this opportunity. If not probably other tool will take over this share.

Xylus:
Hi @dawidmalina 
I'm currently working on accepting and storing UUIDs by encoding them into a 22 byte string.
This should cover your use case in k8s as you would be able to use the pod id as the agent id.
Additionally, since pod id doesn't seem to be very useful to users, we're thinking of collecting the pod name as well and exposing this in the web UI in case of k8s agents.

Please let me know if you have any concerns, or think there are other useful information we should collect for k8s agents.
Thanks.

dawidmalina:
Thanks for sharing this, I appreciate this. I think that pod id for uniqueness and pod name as display name is the best approach 👍 For now I can't find any other useful information that we should collect. Let's start with this and improve later :)

Please let me know if you need help with testing.

Xylus:
@dawidmalina Sounds good!

prajidi-blumeglobal:
Is this issue fixed in branch 1.8.4?

Xylus:
@prajidi-blumeglobal No, it isn't fixed in 1.8.4, and I don't think it will be released for 1.8.x.
We're targetting this for our next non-1.8.x release.

Xylus:
Current plan is to take the 3rd route and keep backwards compatibility - using agent id as an alias, and assigning uuid to each agent id for use internally.
Should involve a lot of code changes to all 3 components so might take a bit to implement.

[...]

PR #6015, opened 2019-09-20
Increasing length limit in variables agentId and applicationName

nightguide:
Added the ability to specify in -Dpinpoint.agentId and -Dpinpoint.applicationName =
names are more than 24 characters.
Now this limit is limited to 100 characters.


emeroad:
Unfortunately, this fix is not valid.

See the following issue.
https://github.com/naver/pinpoint/issues/5343

PR #6055, opened 2019-10-04
[#noissue] fix useless use of cat

dyyim741:
Hello.

I modified the UUOC(useless use of cat)

Pointing out UUOC is a long standing shell programming tradition, and removing them from a short-lived pipeline in a loop can speed it up by 2x. However, it's not necessarily a good use of time in practice, and rarely affects correctness.

what do you think?

Thanks.

Ref: https://en.wikipedia.org/wiki/Cat_(Unix)#Useless_use_of_cat


koo-taejin:
@dyyim741 
I perfectly agree with you're opinion. 
Thanks for you're contribution. :)

Issue #6048, opened 2019-10-02
elasticsearch bboss test failure


RoySRose:
Hello, @yin-bp 

I've noticed that bboss plugin integration tests are failing in our daily routine test.

Do you mind checking `ElasticsearchExecutorIT.java` out?

yin-bp:
Ok, let me look at this problem and solve it

RoySRose:
attaching log, which may help you better
```
Error Message
java.lang.AssertionError: Cannot find apiId of [org.frameworkset.elasticsearch.client.ConfigRestClientUtil.createIndiceMapping(java.lang.String, java.lang.String)]
Stacktrace
com.navercorp.pinpoint.test.plugin.PinpointPluginTestException: java.lang.AssertionError: Cannot find apiId of [org.frameworkset.elasticsearch.client.ConfigRestClientUtil.createIndiceMapping(java.lang.String, java.lang.String)]
	at com.navercorp.pinpoint.plugin.elasticsearchbboss.ElasticsearchExecutorIT.indiceCreate(ElasticsearchExecutorIT.java:116)
	at com.navercorp.pinpoint.plugin.elasticsearchbboss.ElasticsearchExecutorIT.testClientInterface(ElasticsearchExecutorIT.java:101)
```

yin-bp:
Hi,@RoySRose :
This problem is caused by not enabling ElasticsearchBboss plugin, can enable the plugin in pinpoint.config:
profiler.elasticsearchbboss.enabled=true

PR #6059, opened 2019-10-07
Enable Elasticsearch bboss plugin to fix bug #6048

yin-bp:
2.Enable  Elasticsearch bboss plugin to fix bug #6048


RoySRose:
Hello, @yin-bp 

I don't think `log4j.xml` is necessary here. What do you think?

yin-bp:
@RoySRose 
There are some log4j warn when run elasticsearch bboss integration test,so I think here need a log4j.xml file:
```
log4j:WARN No appenders could be found for logger (org.eclipse.aether.internal.impl.DefaultLocalRepositoryProvider).
log4j:WARN Please initialize the log4j system properly.
log4j:WARN See http://logging.apache.org/log4j/1.2/faq.html#noconfig for more info.
```

RoySRose:
Hello, @yin-bp 

Great~!
As usual. can you merge the commits into one instead of two?
since the first commit includes log4j and the second one just removes it again.

yin-bp:
for squashed  and merge commits

PR #6060, opened 2019-10-08
fix agent config comment error

yjqg6666:
Slash style comment is not compatible with jdk property utility.


koo-taejin:
@yjqg6666 thanks for you're contribution :)

PR #6070, opened 2019-10-10
fix missing Mongodb servermap icon for UI v2

yjqg6666:
NO_IMAGE_FOUND.png is used for MongoDB when UI v2 is used.


RoySRose:

PR #6083, opened 2019-10-14
add tcp connect config for profiler

yjqg6666:
* add tcp connect timeout config for span data sender and stat data sender
* add tcp reconnect interval config for span data sender and stat data sender


koo-taejin:
I'm sorry, but many settings have changed, so this pull request is not currently needed.

Issue #6084, opened 2019-10-14
The statement about profiler.sampling.rate in faq.md seems not accurate


AlphaWang:
#### Prerequisites
Please check the [FAQ](http://naver.github.io/pinpoint/faq.html), and search existing [issues](https://github.com/naver/pinpoint/issues) for similar questions before creating a new issue.**YOU MAY DELETE THIS PREREQUISITES SECTION.**

- [x] I have checked the [FAQ](http://naver.github.io/pinpoint/faq.html), and [issues](https://github.com/naver/pinpoint/issues) and found no answer.

In the FAQ document(https://github.com/naver/pinpoint/blob/1.8.4/doc/faq.md), it says `The agent's release binary has this value set to 20`, but actually the value is 1 (https://github.com/naver/pinpoint/blob/1.8.4/agent/src/main/resources-release/pinpoint.config#L49), it may cause some confusion; Is it possible to consider adjusting the wording in faq.md？

Xylus:
Hi @AlphaWang 
You're absolutely right! We have changed the default to 1, but forgot to update the FAQ.
Thank you for pointing this out.

Let me know if you'd like to update the documentation and throw us a PR. If you do decide to do this, please feel free to update the *faq.md* in the master branch as well.
Thanks.

AlphaWang:
Thank you @Xylus for the reply. We may change the doc to a more general statement like `Pinpoint agent samples 1 trace every N transactions if this value was set as N`. I raised 2 PRs for both 1.8.x and master, please help review the wording. Thanks.

PR #6086, opened 2019-10-14
[#6084] reword the statement about profiler.sampling.rate in faq.md

AlphaWang:
Reword the statement about profiler.sampling.rate in faq.md. 
Pinpoint release binary has changed the `profiler.sampling.rate` to 1, but it is still 20 in the document, it might be better to change the doc to a more general statement.


Xylus:
Awesome, thank you for the fix!

PR #6123, opened 2019-10-24
[#noissue] update travis.dist

emeroad:


emeroad:
Xenial does not seem to support jdk6.

Issue #6260, opened 2019-12-03
Update ASM version 7.1 to 7.2


jaehong-kim:
## Pinpoint now supports Java13
Update ASM(https://asm.ow2.io/)

PR #6262, opened 2019-12-03
[#6260] Update ASM version 7.1 to 7.2

jaehong-kim:
#6260

### DIFF
--- a/profiler-optional/profiler-optional-jdk9/src/main/java/com/navercorp/pinpoint/profiler/shutdown/Java9ShutdownHookRegister.java
+++ b/profiler-optional/profiler-optional-jdk9/src/main/java/com/navercorp/pinpoint/profiler/shutdown/Java9ShutdownHookRegister.java
@@ -16,11 +16,13 @@
 
 package com.navercorp.pinpoint.profiler.shutdown;
 
+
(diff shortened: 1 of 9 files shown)


profiler-optional/profiler-optional-jdk9/src/main/java/com/navercorp/pinpoint/profiler/shutdown/Java9ShutdownHookRegister.java
```diff
@@ -16,29 +16,33 @@
...
+
     @Override
     public void register(Thread thread) {
         logger.info("register() started.");
 
-        JavaLangAccess javaLangAccess = SharedSecrets.getJavaLangAccess();
         for (int i = 3; i < 10; i++) {
             try {
-                javaLangAccess.registerShutdownHook(i, true, thread);
+                REGISTER_SHUTDOWN_HOOK_METHOD_HANDLE.invoke(OBJECT, i, true, thread);
```
koo-taejin, line 45: The REGISTER_SHUTDOWN_HOOK_METHOD_HANDLE method seems to be can null. 
It would be better to record the log if it is null.

PR #6322, opened 2019-12-16
fix spanId duplicate in dubbo plugin. since dubbo 2.6.2 it will force will force put RpcContext.attachment to dubbo consumer RpcInvocation

zifeihan:
clear {@link com.alibaba.dubbo.rpc.RpcContext#getAttachments()} trace header.
you should to know,since dubbo 2.6.2 version.
{@link com.alibaba.dubbo.rpc.protocol.AbstractInvoker#invoke(com.alibaba.dubbo.rpc.Invocation)}
will force put {@link com.alibaba.dubbo.rpc.RpcContext#getAttachments()} to current Invocation
replace origin invocation.addAttachmentsIfAbsent(context) method;
to imagine if application(B) methodB called by application(A), application(B) is dubbo provider, methodB call next dubbo application(C).
when application(C) received trace header is overwrite by application(B) received trace header.


zifeihan:
@jaehong-kim here has some checks were not successful, does it influent my pull request?
image

Issue #6456, opened 2020-01-29
Update reactor plugin for scheduler


jaehong-kim:
Add reactor-core scheduler
~~~
            // Scheduler - Provides an abstract asynchronous boundary to operators.
            "reactor.core.scheduler.BoundedElasticScheduler",
            "reactor.core.scheduler.DelegateServiceScheduler",
            "reactor.core.scheduler.SingleScheduler",
            "reactor.core.scheduler.ParallelScheduler",
            "reactor.core.scheduler.ImmediateScheduler",
            "reactor.core.scheduler.ElasticScheduler",
            "reactor.core.scheduler.ExecutorScheduler",
            "reactor.core.scheduler.SingleWorkerScheduler",
            // Worker - A worker representing an asynchronous boundary that executes tasks.
            "reactor.core.scheduler.BoundedElasticScheduler$ActiveWorker",
            "reactor.core.scheduler.BoundedElasticScheduler$DeferredWorker",
            "reactor.core.scheduler.ElasticScheduler$ElasticWorker",
            "reactor.core.scheduler.ExecutorScheduler$ExecutorSchedulerTrampolineWorker",
            "reactor.core.scheduler.ExecutorScheduler$ExecutorSchedulerWorker",
            "reactor.core.scheduler.ExecutorServiceWorker",
            "reactor.core.scheduler.ImmediateScheduler$ImmediateSchedulerWorker",
            "reactor.core.scheduler.SingleWorkerScheduler"

~~~

Issue #6352, opened 2019-12-22
Unable to get call stack in spring-webflux boots on netty server


dinesh4747:
Environment:

Pinpoint Agent version - **2.0.0-RC1**
Spring boot version - **Spring Boot 2.1.3.RELEASE**
App code runs on - **reactor-netty server which will be started inherently since it uses spring webflux**

Issue - Unable to get the scattered point which helps in discovering the 

Below exception indicated that spring boot with reactor netty server which comes with spring webflux 

**2019-12-22 18:46:49 [INFO ](c.n.p.c.r.g.s.c.GrpcCommandService ) checkapp:addorder:1577040406302 => local. ready() transportId:14
2019-12-22 18:46:49 [INFO ](.n.p.c.r.g.s.AgentLifecycleListener) connect:PingSession{id=14, header=Header{agentId='addorder', applicationName='checkapp', agentStartTime=1577040406302, socketId=1}, eventIdAllocator=0}
2019-12-22 18:46:49 [INFO ](c.n.p.c.r.g.s.c.GrpcCommandService ) checkapp:addorder:1577040406302 => local. execute handshake:[740, 710, 730, 750]
2019-12-22 18:46:49 [INFO ](c.n.p.c.r.g.PinpointGrpcServer     ) AgentInfo{checkapp:addorder:1577040406302:} handleHandshake() started. data:[740, 710, 730, 750]
2019-12-22 18:56:54 [INFO ](c.n.p.c.r.g.s.SpanService          ) Failed to span stream, cause=CANCELLED: cancelled before receiving half close
2019-12-22 18:56:54 [INFO ](c.n.p.c.r.g.s.AgentService         ) Failed to ping stream, id=2, cause=CANCELLED: cancelled before receiving half close
2019-12-22 18:56:54 [INFO ](c.n.p.c.r.g.s.StatService          ) Failed to stat stream, cause=CANCELLED: cancelled before receiving half close
2019-12-22 18:56:54 [INFO ](c.n.p.c.r.g.s.c.GrpcCommandService ) Failed to command stream, checkapp:addorder:1577040406302 => local, cause=CANCELLED: cancelled before receiving half close
2019-12-22 18:56:54 [INFO ](.n.p.c.r.g.s.AgentLifecycleListener) close:PingSession{id=14, header=Header{agentId='addorder', applicationName='checkapp', agentStartTime=1577040406302, socketId=1}, eventIdAllocator=3}/ShutdownEventListener{shutdown=false}
[...]

PR #6457, opened 2020-01-29
[#6456] Update reactor plugin for scheduler

jaehong-kim:
#6456


dinesh4747:
@jaehong-kim - Does this help in tracing mono and flux calls , reference to https://github.com/naver/pinpoint/issues/6352

Great if you could confirm the same?

PR #4790, opened 2018-10-24
add traceId search button

linxiaoJVM:
at my project, use MDC feature of log4j and traceId output to log file, the log file store in Elasticsearch, below:

qq 20181024150053

I need search by traceId at pinpoint web, so add a search button at web main html. below:

qq 20181024150524

PR #6543, opened 2020-02-27
add a transactionId header to the response for Jetty&Tomcat

yjqg6666:
1. default to off, no header added to the response.
2. header name can be configured.
3. header value i.e. transactionId is url-encoded by default.

More convenient to check the current specific transaction. 
Enable to trace the transaction for the apps/devices.

To view the transaction, you may check #4790 if you are use V1 UI and
if you are use V2 UI, you may check via the url https://youdomain/v2/transactionDetail/{txId}/0/{agentId}/-1. 
If your txId is jt4zp-10.20.3.4^1581804767659^1, then the agentId is jt4zp-10.20.3.4.


emeroad:
Unfortunately Pinpoint does not use response headers.
I still don't know all the benefits and risks of using a response header.
Why do you want to use response headers?

yjqg6666:
Android or Ios apps call my http api, but the apps can not use java agent. My app may log the txId fetching from the response header. Check the log and find out the txId and view the transaction.

andy263:
useful update for me

alanchenup:
thread-safe problem？ 
image

Issue #6544, opened 2020-02-27
SpanMapperV2#bindSpanChunk contains always true condition


tankilo:
#### Prerequisites
- [x] I have checked the [FAQ](http://naver.github.io/pinpoint/faq.html), and [issues](https://github.com/naver/pinpoint/issues) and found no answer.

#### What version of pinpoint are you using?
master
#### Describe the bug
Instance `spanMap` is contructed in method  `com.navercorp.pinpoint.web.mapper.SpanMapperV2#mapRow` with implementation class as `com.google.common.collect.LinkedListMultimap` and finally passed to method  `com.navercorp.pinpoint.web.mapper.SpanMapperV2#bindSpanChunk` as first argument.

`com.google.common.collect.LinkedListMultimap#get` will never return null.
![image](https://user-images.githubusercontent.com/10413284/75430585-7ca68900-5986-11ea-9aaf-75b44a192b99.png)

![image](https://user-images.githubusercontent.com/10413284/75430935-0f472800-5987-11ea-8c0a-ecdb5cd288b4.png)

#### Expected behavior
`if (matchedSpanBoList != null) {` replaced with `if (!CollectionUtils.isEmpty(matchedSpanBoList)) {`

emeroad:
Thank you for your contribution
I fully agree with you.

PR #6545, opened 2020-02-27
#6544 SpanMapperV2#bindSpanChunk contains always true condition

tankilo:


tankilo:
> Thank you for your contribution
> 
> Would you please remove the !(NOT) operator?
> 
> ```java
> if (CollectionUtils.hasLength(matchedSpanBoList)) {
> ~~
> }
> ```

You are right.

PR #6634, opened 2020-03-24
New feature Informix JDBC plugin

guillermomolina:
This is a plugin for Informix JDBC. 

It is based on the other plugins.
I am using codes for service type:
2450 INFORMIX
2451 INFORMIX_EXECUTE_QUERY

I couldn't add to the tests:

Because I don't know what all the numbers came from.
commons-server/src/test/java/com/navercorp/pinpoint/common/server/bo/codec/stat/join/DataSourceCodecTest.java
commons-server/src/test/java/com/navercorp/pinpoint/common/server/bo/stat/join/JoinApplicationStatBoTest.java
commons-server/src/test/java/com/navercorp/pinpoint/common/server/bo/stat/join/JoinDataSourceListBoTest.java
web/src/test/java/com/navercorp/pinpoint/web/mapper/stat/sampling/sampler/JoinDataSourceSamplerTest.java


emeroad:

Refer to msql-jdbc-it module for testing
 - https://github.com/naver/pinpoint/tree/master/plugins-it/mssql-jdbc-it

commons/src/main/java/com/navercorp/pinpoint/common/trace/ServiceType.java
```diff
@@ -98,6 +98,8 @@
  * <tr><td>2301</td><td>ORACLE_EXECUTE_QUERY</td></tr>
  * <tr><td>2400</td><td>CUBRID</td></tr>
  * <tr><td>2401</td><td>CUBRID_EXECUTE_QUERY</td></tr>
+ * <tr><td>2400</td><td>INFORMIX</td></tr>
```
koo-taejin: Please check servicetype again.

koo-taejin:
hi @guillermomolina 

First of all, I really appreciate for your contribution.
I have been reviewing your code. 

Please check below things. 
`com.informix.jdbc.IfxSqliConnect#IfxSqliConnect(java.lang.String, java.lang.String, java.lang.String, java.util.Properties) `

It looks like you can create DatabaseInfo object through one of IfxSqliConnect constructor parameter.
I think if you use it, you don't need to create unnecessary Getter and you can use common code. 

If it is possible to use the constructor's parameters, please change it to use them.

If I have misunderstood, please feel free to tell me it.

thanks :)

guillermomolina:
Hi @kao-taejin,
Thank you for reviewing my code.

Indeed, at first I was just using the constructor, but that only works for direct connections (without pools). At least in our production environment, most of the connections are made through the pool, so I ended up adding the getter because the database name and connection info is not passed to the constructor. It is somehow gotten from the pool. So, as I had to create the getters anyway for the pooled connections, I am always doing it this way.

Does it sounds right to you?

Thanks

koo-taejin:
@guillermomolina 

I fully understood the situation through the explanation.

thanks :)

plugins/informix-jdbc/src/main/java/com/navercorp/pinpoint/plugin/jdbc/informix/interceptor/InformixStatementCreateInterceptor.java
```diff
@@ -0,0 +1,117 @@
...
+            return;
+        }
+        final DatabaseInfo databaseInfo = getDatabaseInfo(target);
+        if (result instanceof DatabaseInfoAccessor) {
+            ((DatabaseInfoAccessor) result)._$PINPOINT$_setDatabaseInfo(databaseInfo);
+        }
+        SpanEventRecorder recorder = trace.currentSpanEventRecorder();
+        recorder.recordServiceType(databaseInfo.getType());
+        recorder.recordEndPoint(databaseInfo.getMultipleHost());
+        recorder.recordDestinationId(databaseInfo.getDatabaseId());
```
koo-taejin: In general, the role of the StateCreateInterceptor is to attach the DatabaseInfo object for future use.
So, does not need to execute method of recording, because does not track at here. 
I wonder you have some reason for this, otherwise please remove it.

guillermomolina: This is related to the previous question. It is at this point that I can query the database info, that it's otherwise invalid. But I don't really fully understand pinpoint internals, so I there is another way of doing it, please let me know

koo-taejin: I checked it again. 
Honestly, I don't seem to need this.
If you look at other JDBC driver plugins, there is such a process. (They all do this in the Connection constructor.)
However, in this case, it is for augmenting the tracking data when creating a connection.

However, creating a Connection within the createStatement method or augmenting this data seems unusual, so I think this is unnecessary.
If you don't have this logic, I'm wondering if the look you're tracking will look different.

guillermomolina: I will check it out

guillermomolina: Ok, I've tested your proposal. You were right, It does work ok. I didn't understand you at first. I though you want me to kill the entire interceptor, not just the recording part. I show you prove that it still works:

After

koo-taejin:
@guillermomolina 

Everything looks good. 
If you make the commits as one, I will merge it immediately.

thanks :)

koo-taejin:
@guillermomolina 👍

koo-taejin:
@guillermomolina 
I have been testing about the InformixPlugin. 

I found out critical issue.
As of the latest version informix 4.50.x, the field type of connInfo has been changed from java.util.Properties to com.informix.util.AdvancedUppercaseProperties, so the plugin cannot be used in the 4.50.x version.

Would it be possible to ask of you solving this problem?

guillermomolina:
> @guillermomolina
> I have been testing about the InformixPlugin.
> 
> I found out critical issue.
> As of the latest version informix 4.50.x, the field type of connInfo has been changed from java.util.Properties to com.informix.util.AdvancedUppercaseProperties, so the plugin cannot be used in the 4.50.x version.
> 
> Would it be possible to ask of you solving this problem?

I will look into it

Issue #6657, opened 2020-03-30
Update network test to cover grpc


RoySRose:
adding grpc ports in tools/network checker

PR #6658, opened 2020-03-30
[#6657] update test to check grpc ports

RoySRose:


tools/src/main/java/com/navercorp/pinpoint/tools/NetworkAvailabilityChecker.java
```diff
@@ -57,24 +56,94 @@ public static void main(String[] args) {
...
+
+    private static void checkGRPCMeta(GrpcTransportConfig grpcTransportConfig) throws Exception {
+        String ip = grpcTransportConfig.getMetadataCollectorIp();
+        int port = grpcTransportConfig.getMetadataCollectorPort();
+
+        NetworkChecker checker = new TCPChecker("TCP Meta", ip, port);
+        checker.check();
+    }
+
+    private static void checkGRPCStaT(GrpcTransportConfig grpcTransportConfig) throws Exception {
```
koo-taejin, line 133: I think "checkGRPCStat" is better than "checkGRPCStaT" :)

RoySRose: got it

PR #6754, opened 2020-04-30
[#6760] add system property config support

yjqg6666:
System property take higher priority than the config file for the specified profile.
From the source code https://github.com/naver/pinpoint/blob/c5e6b7e33343ad8122823600f5977ba8927b6fe6/bootstrap/src/main/java/com/navercorp/pinpoint/bootstrap/config/ProfilePropertyLoader.java#L76, i see the comment  and the feature not implemented. Should i consider something additional for that?


emeroad:
@yjqg6666 

It's not a good idea to apply all keys, so I didn't implement this feature

https://github.com/naver/pinpoint/pull/6754/files#diff-65ff9c58feaada58b6781ccc504168fbR93
```JAVA
for (String propertyName : stringPropertyNames) {
            boolean isPinpointProperty = propertyName.startsWith("bytecode.") || propertyName.startsWith("profiler.") || propertyName.startsWith("pinpoint.");
            if (isPinpointProperty) {
            ~~
      
}

```
This filter code is good, so there seems to be no problem.

PR #6769, opened 2020-05-07
[#6790] Support filter transactions on App node

yjqg6666:
Context-click on application node, "Filter Transaction Wizard" to filter transactions without restriction of from/to specific node. Similar to the "Filter Transaction Wizard" of node-to-node.


emeroad:
@yjqg6666
Wow~ It's cool
This feature seems very useful
I will review it next week.

PR #6775, opened 2020-05-09
Update transaction detail view in web.

yjqg6666:
- In web ui transaction detail view, put application name to the first column by default for quicker navigation.
before:
image

after:
image

- Update style for transaction short info container to show some hide characters.
before:
image
after:
image


binDongKim:
@yjqg6666 been occupied with other issues lately but I'll review your PRs asap. Thanks for your many contribution.

binDongKim:
@yjqg6666 We've discussed this and decided to leave the "Application" column behind as it was because the unnamed color column in the first column helps users to notice the boundary. So, I changed your commit, just accepting the updating style. Thank you.

PR #6784, opened 2020-05-12
Thread plugin support tracing multiple packages.

yjqg6666:
- Feature
The config profiler.thread.match.package can now be configured with multiple package names, instead of one package name which includes all subpackages.
- Why do we need this feature?
Current profiler.thread.match.package config entity only support one package. If we want to trace com.example.bizA.async and com.example.bizB.async, the config value can only be com.example(which may matches too many packages, com.example.bizC would be included).

### DIFF
--- a/plugins/thread/src/main/java/com/navercorp/pinpoint/plugin/thread/ThreadPlugin.java
+++ b/plugins/thread/src/main/java/com/navercorp/pinpoint/plugin/thread/ThreadPlugin.java
@@ -45,11 +45,8 @@ public class ThreadPlugin implements ProfilerPlugin, MatchableTransformTemplateA
         }
         List<String> threadMatchPackageList = StringUtils.tokenizeToStringList(threadMatchPackages, ",");
         for (String threadMatchPackage : threadMatchPackageList) {
-            String trimmedMatchPackage = threadMatchPackage.trim();
-            if(trimmedMatchPackage.length() > 0) {
-                addRunnableInterceptor(trimmedMatchPackage);
-                addCallableInterceptor(trimmedMatchPackage);
-            }
+            addRunnableInterceptor(threadMatchPackage);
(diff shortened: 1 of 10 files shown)


yjqg6666:
The IT test classes are moved to packages under com.navercorp.test.pinpoint.plugin.thread. They were under com.navercorp.test.pinpoint.plugin.

plugins/thread/src/main/java/com/navercorp/pinpoint/plugin/thread/ThreadPlugin.java
```diff
@@ -38,16 +38,23 @@ public void setup(ProfilerPluginSetupContext context) {
...
+        String threadMatchPackages = threadConfig.getThreadMatchPackage();
+        if (StringUtils.isEmpty(threadMatchPackages)) {
             logger.info("thread plugin package is empty,skip it");
             return;
         }
-        addRunnableInterceptor(threadConfig);
-        addCallableInterceptor(threadConfig);
+        String[] threadMatchPackageList = threadMatchPackages.split(",");
+        for (String threadMatchPackage : threadMatchPackageList) {
+            String trimmedMatchPackage = threadMatchPackage.trim();
```
emeroad: @yjqg6666
https://github.com/naver/pinpoint/blob/master/commons/src/main/java/com/navercorp/pinpoint/common/util/StringUtils.java#L141

Can you change to `StringUtils.tokenizeToStringList()` API

yjqg6666: Updated & rebased & force-pushed.

yjqg6666: @emeroad

PR #6789, opened 2020-05-14
append git.commit.id.abbrev in build.info to agentVersion

yjqg6666:
Parse build.info property file in agent directory, append the build git commit id abbreviation to the agent version of agent information if no error on parsing. It's quite useful for self-build&customized agent.
Before:
image

After:
image


yjqg6666:
We could also append extension version to git.commit.id.abbrev in build.info, ie. 3ed4xxx.ExtV18.

PR #6801, opened 2020-05-18
support export trace info to app for external usage

yjqg6666:
# Feature
Export txId & spanId to application. Disabled by default.
# Usage
If my application need to save/notify some system the txId/spanId, i could 
1) import         com.navercorp.pinpoint:pinpoint-application-interaction:2.0.2 as dependency.
2)  fetch the txId/spanId

> TraceInfo traceInfo = TraceInfoHolder.getTraceInfo();  
> String txId = TraceInfoHolder.getTransactionId();  
> long spanId = TraceInfoHolder.getSpanId();  

# Scene
In kafka lower version  client/broker which do not support KAFKA header, now we could get the txId & spanId and put these info into message body, then the consumer can get the trace info and log. Now i can also save txId & my biz id(order id/user id etc) to db together.

Before this PR, we could only get the txId/spanId from slf4j.MDC in a hacky way.


emeroad:
@yjqg6666 
I have a question
If I remove pinpiont-agent in an application using export api, does it work normally?

yjqg6666:
> @yjqg6666
> I have a question
> If I remove pinpiont-agent in an application using export api, does it work normally?

It would not affect the application, just no write only read, null TraceInfo/TransactionId would be return.

yjqg6666:
> TraceInfo traceInfo = TraceInfoHolder.getTraceInfo();
> String txId = TraceInfoHolder.getTransactionId();
> long spanId = TraceInfoHolder.getSpanId();
If no pinpoint in javaagent or the request is not sampled, the above three variables would always be null.  The package com.navercorp.pinpoint:pinpoint-application-interaction must have no extra dependency introduced.

Issue #6809, opened 2020-05-21
Support the function of storing custom metrics


koo-taejin:
#### description 
At the current pinpoint, each data in Inspector page are based on the data that most of the agent have.
Develops new feature to enable the storage of information about metrics specific to individual agents.
It is expected that this feature will help user finds out problems

koo-taejin:
Due to considering to apply OLAP system, the work has been postponed.

PR #6810, opened 2020-05-21
[#6809] Support the function of storing custom metrics

koo-taejin:
This feature is under devlopement.

* commit description
 - Allows agents to collect CustomMetrics. (After collecting, CustomMetric will be discarded without being transmitted.)


koo-taejin:
Some codes have changed via https://github.com/naver/pinpoint/commit/77d244feca0c554fec6f16d4b24cadf09a8db130

- to create new thread for sending custom metric -> to use exist thread that have sent stat data.
- to allow all of custom metrics -> to allow specific custom metric
- change CustomMetric classname.

PR #6823, opened 2020-05-26
Always record request having Pinpoint-Force-Sample request header

yjqg6666:
- If the http request have Pinpoint-Force-Sample request header and the request path is not excluded, always record the request regardless of sampling rate.
- This feature is off by default for the security reason & backward compatibility, can be enabled by profiler.sampling.force.header=true.  We could turn this feature on in development environment(dev profile) and off in production environment(release profile).

### DIFF
--- a/bootstrap-core/src/main/java/com/navercorp/pinpoint/bootstrap/context/Header.java
+++ b/bootstrap-core/src/main/java/com/navercorp/pinpoint/bootstrap/context/Header.java
@@ -28,6 +28,7 @@ public enum Header {
     HTTP_SPAN_ID("Pinpoint-SpanID"),
     HTTP_PARENT_SPAN_ID("Pinpoint-pSpanID"),
     HTTP_SAMPLED("Pinpoint-Sampled"),
+    HTTP_FORCE_SAMPLE("Pinpoint-Force-Sample"),
     HTTP_FLAGS("Pinpoint-Flags"),
     HTTP_PARENT_APPLICATION_NAME("Pinpoint-pAppName"),
     HTTP_PARENT_APPLICATION_TYPE("Pinpoint-pAppType"),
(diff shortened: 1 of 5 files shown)


emeroad:
@yjqg6666 
Hi, I agree with your opinion.
I think pinpoint needs a debugging header for the development environment.

However, it is difficult to change Sampler in the current API.
First of all, I plan to add an API for sampling policy.

emeroad:
I'm thinking of two ways

## Method 1
Add manual API to TraceContext.
  -TraceContext::newForceTraceObject () Trace
  -TraceContext::newDisableTraceObject () Trace

## Method 2
Add a new Sampler API to understand RequestContext.

```java
public interface ReqeustContextSampler {
       boolean isSampling (RequestAdaptor <T>, T);
}


public class DebuggingContextSampler implements ReqeustContextSampler {
       boolean isSampling (RequestAdaptor <T> ra, T v) {
               if (ra.getHeader ("Pinpoint-Debugging-Sample ", v) == "true" ) {
                       return true;
               }
       }
}

public class TraceContext {
     ~~~
      // Add new API
     Trace newTraceObject(RequestAdaptor<T> ra, T v);

     // Trace newTraceObject(RequestAdaptor<T> ra, T v, Options extendedOption);
}

```
I think it's good to add a new Sampler API, but I didn't have enough time to try.
Could you implement it?

yjqg6666:
@emeroad Actually our usage for this feature would be that if the controller have some specific annotations, the request calling the controller would be always recorded regardless of environment. 
We would scan the annotations before servicing and put them into a list,  when the request come, pinpoint intercept the request and check for tracing, if true then force tracing. 
Recent i may not have time for that，i have to complete the rest.

yjqg6666:
If using the method 1, the problem would be we have to add newForceAsyncTraceObject, newForceTraceObject, newDisableTraceObject. Agree with you. 
https://github.com/naver/pinpoint/blob/ce0c257122a621f64d578b6258627c7d75c91333/bootstrap-core/src/main/java/com/navercorp/pinpoint/bootstrap/plugin/request/RequestTraceReader.java#L132-L137

Prefer method 2, we can decide whether sampling or not based on request, ex. request header, request path, request host, request cookie.

RoySRose:
> Actually our usage for this feature would be that if the controller have some specific annotations, the request calling the controller would be always recorded regardless of environment.

I think this is a great feature

yjqg6666:
@emeroad @RoySRose I make another PR #6834 for this.  We may refactor the internal implementation, now it's just working as expected.

PR #6837, opened 2020-05-29
[Web] show endpoint(host+port) in rpc.argument

yjqg6666:
Before:
image
After:
image


emeroad:
The endpoint information seems useful.
I added this as a separate record.

https://github.com/naver/pinpoint/pull/6844

yjqg6666:
image
A new column could also be added, the transactionmetadata api has already returned the endpoint data.
@emeroad

PR #6854, opened 2020-06-02
fix esc to clear search txId input box

yjqg6666:
Fix press escape key won't clear the search box.

### DIFF
--- a/web/src/main/angular/src/app/core/components/transaction-id-search/transaction-id-search-container.component.html
+++ b/web/src/main/angular/src/app/core/components/transaction-id-search/transaction-id-search-container.component.html
@@ -6,6 +6,7 @@
         class="l-search-input"
         [useEnter]="searchUseEnter"
         (outSearch)="onSearchId($event)"
+        (keydown.esc)="onClearSearch(searchInput)"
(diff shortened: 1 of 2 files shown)


yjqg6666:
@binDongKim

binDongKim:
Esc key clearing should already work with the `ppSearchInput` directive because it's already implemented there. It works in my local.

yjqg6666:
Not working in my fresh build from master. Press ESC key only get the focus lost.
image

yjqg6666:
It's also not working in an app's page.
image

yjqg6666:
I found out the reason. If i disable the cVim plugin on the domain (in Chrome), it's working. 
I wonder why "(keydown.esc)" is working while ppSearchInput don't.

yjqg6666:
cVim is not available on Chrome Web Store, but can be downloaded from https://github.com/1995eaton/chromium-vim.

binDongKim:
@yjqg6666 I'll look into it. Thanks for the report.

binDongKim:
@yjqg6666 I'll close this and dig into the relation between the cVim and angular later.

Issue #6745, opened 2020-04-27
make netty direct buffer memory metric


yangbongsoo:
``` java
boolean directBufferPreferred = io.netty.util.internal.PlatformDependent.directBufferPreferred();
long usedDirectMemory = io.netty.util.internal.PlatformDependent.usedDirectMemory();
long maxDirectMemory = io.netty.util.internal.PlatformDependent.maxDirectMemory();
```

cf) if below option is true, directBufferPreferred value is false.
```
-Dio.netty.noPreferDirect=true
```

koo-taejin:
hi @yangbongsoo 

It is impossible to develop this function under the present situation.
Each data in Inspector page are based on the data that most of the agent have.
Data that this function will show is depended on Netty. 
Therefore, Pinpoint needs to develop additional features to support this function.

I will give you the instructions you need to create it, after creating a new feature.

thanks :)

koo-taejin:
@yangbongsoo 

Please record the data that needs to be collected periodically as follows.

I am not sure how to make it ideal for implementation.
So, I let records save with permisson.

I have defined two AllowedSources for you.
```
    private static final AllowedSource<LongCounter> NETTY_USED_DIRECT_MEMORY = new AllowedSource<LongCounter>("custom/netty/usedDirectMemory", LongCounter.class);
    private static final AllowedSource<LongCounter> NETTY_MAX_DIRECT_MEMORY = new AllowedSource<LongCounter>("custom/netty/maxDirectMemory ", LongCounter.class);
```


### Registration
Add `CustomMetricRegistry customMetricMonitorRegistry` in  constructor of Interceptor. 

Ex) 
* before
```
    public StandardHostValveInvokeInterceptor(TraceContext traceContext, MethodDescriptor descriptor, RequestRecorderFactory<HttpServletRequest> requestRecorderFactory) {
```

* after
```
    public StandardHostValveInvokeInterceptor(TraceContext traceContext, MethodDescriptor descriptor, RequestRecorderFactory<HttpServletRequest> requestRecorderFactory, CustomMetricRegistry customMetricMonitorRegistry) {
```

### Implemetation
Include a method that periodically returns the value that needs to be extracted by implementing the LongCounter interface.

Ex)
```

private UsedDirectMemoryCounter implements LongCounter  {

        private final String name;

        public UsedDirectMemoryCounter(String name) {
            this.name = name;
        }

        @Override
        public long getValue() {
		    // to make your own code.
        }

        @Override
[...]

PR #6975, opened 2020-07-03
make netty direct buffer memory metric

yangbongsoo:
#6745

### DIFF
--- a/plugins/netty/src/main/java/com/navercorp/pinpoint/plugin/netty/NettyPlugin.java
+++ b/plugins/netty/src/main/java/com/navercorp/pinpoint/plugin/netty/NettyPlugin.java
@@ -82,7 +82,6 @@ public class NettyPlugin implements ProfilerPlugin, TransformTemplateAware {
             transformTemplate.transform("io.netty.handler.codec.http.HttpObjectEncoder", HttpEncoderTransformer.class);
         }
 
-        // todo
         transformTemplate.transform("io.netty.util.internal.PlatformDependent", PlatformDependentTransform.class);
     }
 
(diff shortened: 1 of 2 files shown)


plugins/netty/src/main/java/com/navercorp/pinpoint/plugin/netty/NettyPlugin.java
```diff
@@ -79,13 +81,28 @@ public void setup(ProfilerPluginSetupContext context) {
...
         this.transformTemplate = transformTemplate;
     }
 
+    public static class PlatformDependentTransform implements TransformCallback {
+        @Override
+        public byte[] doInTransform(Instrumentor instrumentor, ClassLoader loader, String className, Class<?> classBeingRedefined, ProtectionDomain protectionDomain, byte[] classfileBuffer) throws InstrumentException {
+            InstrumentClass target = instrumentor.getInstrumentClass(loader, className, classfileBuffer);
+
+            for (InstrumentMethod method : target.getDeclaredMethods(MethodFilters.name("usedDirectMemory", "maxDirectMemory"))) {
+                method.addInterceptor(PlatformDependentInterceptor.class);
```
yangbongsoo, line 99: @koo-taejin I think that's not necessary.
because we don't intercept when client(I mean agent server) call `usedDirectMemory ` and `maxDirectMemory ` method.

koo-taejin: @yangbongsoo
I think that part is good.
If api removed, then this part can avoid unexpected situation.
(I don't think this api will be removed :)

koo-taejin:
@yangbongsoo 
Please add license at `PlatformDependentInterceptor.java`

PR #6978, opened 2020-07-03
fix period-selector maxPeriodTime issue

yjqg6666:
1. Change the MAX_SEARCH_PERIOD hint message from days unit to hours unit. Please verify that the I18n text(hours= 시간) is correct. Using hours may be more convenient.
2. The maxPeriod value could be read from getMaxPeriodTime config, it was the last value of periodList.

### DIFF
--- a/web/src/main/angular/src/app/core/components/period-selector/period-selector-container.component.ts
+++ b/web/src/main/angular/src/app/core/components/period-selector/period-selector-container.component.ts
@@ -51,7 +51,7 @@ export class PeriodSelectorContainerComponent implements OnInit, OnDestroy {
 
     ngOnInit() {
         this.periodList = this.webAppSettingDataService.getPeriodList(this.newUrlStateNotificationService.getStartPath());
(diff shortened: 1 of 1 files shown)


binDongKim:
What made you think using/displaying the hint message with hours is more convenient? As you can see from the interface, the period-list is listed as `5m / 20m / ... / 2d ` so the hint message with `2d` seems consistent in my opinion.

yjqg6666:
In production environment, we may restrict to use too long time span/period, 2 days or more may cause many hbase lookups, hours may be preferred.

yjqg6666:
If the maxPeriodTime is 6 hours, the message would be 0.25 day. If 3 hours, then 0.125 day.

binDongKim:
I don't understand. The maxPeriodTime has set as `2days` as a static value(or `7days` in inspector page) so the message would be never anything other than `2days`(or `7days`).

yjqg6666:
@binDongKim Updated the PR, only include "honor the getMaxPeriodTime static config".

PR #6996, opened 2020-07-09
Clarify milliseconds is a 13 digit number

jbnimble:
The 10 digit example did not work, and after research found that an epoch in milliseconds is 13 digits, updated the documentation to reflect a 13 digit example


emeroad:
@jbnimble
Thank for you for your contribution

PR #7039, opened 2020-07-21
Fix DB URL null:null

guillermomolina:
InformixPreparedStatementCreateInterceptor now tries to access databaseInfo from the previously recorded info, and creating new info when this info is not available. Before the patch, It always tried to create the info from the class, but sometimes the URL was not available, leaving a connection to null:null.

Screenshot_20200721_123620


emeroad:
@guillermomolina 
Thank you all for your contribution

PR #7070, opened 2020-08-01
[#noissue] typofix (LIst -> List)

Jaeyo:
Maybe some of you like `LIst` instead of `List` 😁


RoySRose:
Hello, @Jaeyo 

Thank you for your first-time contribution.
I agree with you on fixing the typo.

Just like to ask you.
Have you tested it after fixing the names?

Jaeyo:
@RoySRose just ran unit tests by executing `./mvnw test` and it reports success.

RoySRose:
As I've commented same in another PR (https://github.com/naver/pinpoint/pull/7071)

The test might not represent that it is okay since we are not sure that the test for the changes even exists.

I've also looked into the changes and doubt that it will cause any problem.
But highly recommend that you try running the full application.
I believe you need to set the environment anyway.

RoySRose:
Please change the header of the commit and add [#noissue] as below

image

emeroad:

PR #7077, opened 2020-08-04
when profiler.mongodb enabled, only transform mongodb v3 classes

yjqg6666:
When profiler.mongodb enabled, if a project is using mongo-java-driver v2.x, transform exception would be thrown.
Add version check and only profile if the mongodb version is 3.x.

Changes:
1. move all the transform callbacks from MongoPlugin to a new class MongoTransforms.java
2. all the transform callbacks extends AbstractMongoTransformCallback which has version check
3.  all the transform callbacks call the version check before actual transformation.

### DIFF
--- a/plugins-it/mongodb-it/pom.xml
+++ b/plugins-it/mongodb-it/pom.xml
@@ -48,7 +48,8 @@
         </dependency>
         <dependency>
             <groupId>org.mongodb</groupId>
-            <artifactId>mongodb-driver</artifactId>
+            <artifactId>mongo-java-driver</artifactId>
+            <version>3.12.6</version>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/plugins-it/mongodb-it/src/test/java/com/navercorp/pinpoint/plugin/mongodb/MongoDBIT_3_0_x_IT.java
+++ b/plugins-it/mongodb-it/src/test/java/com/navercorp/pinpoint/plugin/mongodb/MongoDBIT_3_0_x_IT.java
@@ -36,7 +36,7 @@ import org.junit.runner.RunWith;
 @JvmVersion(8)
 @ImportPlugin({"com.navercorp.pinpoint:pinpoint-mongodb-driver-plugin"})
 @Dependency({
-        "org.mongodb:mongodb-driver:[3.0.0,3.1.max]",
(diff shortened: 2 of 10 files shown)


emeroad:
@yjqg6666 
Thank your for your contribution 

I plan to add a feature to identify the version in the plugin.
This feature can separate plugins for each version of lib.
It has the advantage of making compilation and maintenance easy.
e.g) mongo-db-plugin-2.x.jar, mongo-db-plugin-3.x.jar, 

```java
// pseudocode
public class MongoPlugin implements ProfilerPlugin, TransformTemplateAware {
    context.addVerionMatcher(new VersionMatcher() {
            @Override
            public boolean isMatch(JarFile file) {
                // Manifest base matcher
                Manifest manifest = file.getManifest();
                Attributes attributes = manifest.getAttributes("PATH");
                String version = attributes.getValue(Attributes.Name.IMPLEMENTATION_VERSION);
                if (version.endsWith("2.0.0")) {
                    //~~
                    return true;
                }

                // File name base matcher
                if (file.getName().endsWith("2.0.0")) {
                    //~~
                    return true;
                }
                return true;
            }
        });

}
```

Jarfile can be obtained from ProtectionDomain of Transform API.
```java
// pseudocode
@Override
public byte[] doInTransform(Instrumentor instrumentor, ClassLoader loader, String className, Class<?> classBeingRedefined, ProtectionDomain protectionDomain, byte[] classfileBuffer) throws InstrumentException {
     URL location = protectionDomain.getCodeSource().getLocation();
        if (!location.getProtocol("jar")) {
            //~~
        }
        JarFile jarFile = new JarFile(location.getFile());
        if (!target.isInterceptable()) {
            return null;
        }
}
```

What do you think of this feature?

yjqg6666:
For fat jar, there would be no Manifest file, getPackage().getImplementationVersion() would be null. You could add this feature/util which is awesome, after that this version check part could be replaced.

yjqg6666:
Plus feature, intercept DBCollection & DB(mongodb v3.x only) which may still be used in some applications even though they are deprecated(still exist in v3).

cedrices:
Hi! @emeroad  Has this problem been solved in version 2.2.2？Thanks for you
reply.

cedrices:
Hi! @emeroad  Has this problem been solved in version 2.2.2？Thanks for you
reply.

PR #7157, opened 2020-08-23
[#noissue] Add AlarmWriter Isolation Test

cwJohnPark:
case 1 : when SequenceCount is less than TimingCount, Send alarm
case 2 : when SequenceCount is equal to TimingCount, Don't send alarm


minwoo-jung:
@cwJohnPark 
good! I will review soon.

Issue #6902, opened 2020-06-11
when some method called exception, pinpoint desnot show more  printstack


chenglin321:
#### Prerequisites
Please check the [FAQ](http://naver.github.io/pinpoint/faq.html), and search existing [issues](https://github.com/naver/pinpoint/issues) for similar questions before creating a new issue.**YOU MAY DELETE THIS PREREQUISITES SECTION.**

- [ ] I have checked the [FAQ](http://naver.github.io/pinpoint/faq.html), and [issues](https://github.com/naver/pinpoint/issues) and found no answer.

#### What version of pinpoint are you using?
2.0.2

#### Describe the bug
when some method called exception, pinpoint desnot show more  printstack

![3A698839-C0A7-4338-8AD1-5350BF582BC1](https://user-images.githubusercontent.com/3918176/84350326-af9a2300-abeb-11ea-9e4c-7ba83b49daac.png)

The content has less printstack ,but i want more printstack information

nested exception is org.apache.ibatis.exceptions.PersistenceException: ### Error querying database. Cause: java.lang.IllegalArgumentException: Mapped Statements collection does not contain value for s_i18n.selectCountByParams ### Cause: java.lang.Illegal...(357)

emeroad:
Too long strings that are too long are being truncated
Stacktrace will be supported in the future

PR #7158, opened 2020-08-23

IluckySi:
**Pinpoint** is a very good APM platform, it can help us quickly locate many potential performance problems of application, such as operating MySQL exception, accessing other services timeout and so on. During one use, I found that there was no detailed stack information for the exceptions collected by the agent. In the issue, it was also mentioned that: **https://github.com/naver/pinpoint/issues/6902**
However, sometimes the detailed exception stack is also very important for us to quickly locate the problem, so I added  a featuer, which includes the following two configuration items:
**profiler.exception.stacktrace .enable=false  # default is false
profiler.exception.stacktrace .line=20 # default is 20**

In terms of code, I added a SpanThrowable class to encapsulate Throwable. That is, when applying after logic of interceptor, throwable will be processed. The processing logic is as follows:
1. Judge whether the function 'profiler.exception.stacktrace.enable' is enabled. If not, the detailed exception stack will not be collected;
2. If enabled, collect by  'profiler.exception.stacktrace.line'

Note: most plugins are involved in this adjustment. Plugins that support asynchronous context are not supported for the time being.
Take MySQL as an example，the detailed test process is as follows:
profiler.exception.stacktrace.enable=false
profiler.exception.stacktrace.line=20
image
profiler.exception.stacktrace.enable=true
profiler.exception.stacktrace.line=5
image
profiler.exception.stacktrace.enable=true
profiler.exception.stacktrace.line=-1
image

**Other plugin test:**
profiler.exception.stacktrace.enable=true
profiler.exception.stacktrace.line=5
redis:
image
httpclient:
image
profiler.exception.stacktrace.enable=true
profiler.exception.stacktrace.line=-1
redis:
image
httpclient:
image


IluckySi:
related to https://github.com/naver/pinpoint/pull/7165

PR #7165, opened 2020-08-24

IluckySi:
related to https://github.com/naver/pinpoint/pull/7158

### DIFF
--- a/agent/src/main/resources/pinpoint-root.config
+++ b/agent/src/main/resources/pinpoint-root.config
@@ -233,6 +233,11 @@ profiler.callstack.max.depth=64
 # weather or not to propagate exceptions occurred at interceptor
 profiler.interceptor.exception.propagate=false
 
+## Trace Exception StackTrace
+# Set line, default is 20, if -1 is unlimited.
+profiler.exception.stacktrace.enable=false
+profiler.exception.stacktrace.line=20
+
 # Allow bytecode framework (ASM only)
 profiler.instrument.engine=ASM
 
--- a/bootstrap-core/src/main/java/com/navercorp/pinpoint/bootstrap/config/DefaultProfilerConfig.java
+++ b/bootstrap-core/src/main/java/com/navercorp/pinpoint/bootstrap/config/DefaultProfilerConfig.java
@@ -169,6 +169,9 @@ public class DefaultProfilerConfig implements ProfilerConfig {
     private boolean customMetricEnable = false;
     private int customMetricLimitSize = 10;
 
+    private boolean exceptionStackTraceEnable = false;
+    private int exceptionStackTraceLine = 20;
+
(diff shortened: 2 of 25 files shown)


IluckySi:
@emeroad Hello， Woonduk， I need you to see if there is any problem with this pr，Thank you~
detailed description：https://github.com/naver/pinpoint/pull/7158

emeroad:
@IluckySi 

https://github.com/naver/pinpoint/pull/7165/files#diff-77f872e60c9d162a6ff1fed372a91cceR26

What do you think of migrating `XXXInterceptor.processThrowable` method to `AbstractRecorder.recordException` method?

IluckySi:
> @IluckySi
> Thank you for your contribution.
> 
> https://github.com/naver/pinpoint/pull/7165/files#diff-77f872e60c9d162a6ff1fed372a91cceR26
> 
> What do you think of migrating `XXXInterceptor.processThrowable` method to `AbstractRecorder.recordException` method?

I thought so before, but I found that I cannot get Profilerconfig  in AbstractRecorder.recordException method~

IluckySi:
@emeroad 
What do you think of Changing DefaultProfilerConfig to a SingleTon，
So I can use DefaultProfilerConfig in AbstractRecorder.recordException method，
Do you have any suggestions，Thank you~

` public static DefaultProfilerConfig getInstance () {
        return getInstance(null);
    }

    public static DefaultProfilerConfig getInstance (Properties properties) {
        if (defaultProfilerConfig == null) {
            synchronized(DefaultProfilerConfig.class) {
                if (defaultProfilerConfig == null) {
                    defaultProfilerConfig = new DefaultProfilerConfig(properties);
                }
            }
        }
        return defaultProfilerConfig;
    }

    // For test
    public DefaultProfilerConfig() {
        this.properties = new Properties();
        this.thriftTransportConfig = new DefaultThriftTransportConfig();
    }

    // For test
    public DefaultProfilerConfig(Properties properties) {
        if (properties == null) {
            throw new NullPointerException("properties");
        }
        this.properties = properties;
        readPropertyValues();
    }`

tankilo:
Should the root cause of the exception to be record before outside exception?
Sometimes, exception may be throwed from inner methods which are not intercepted and then intercepted by outer methods, the root cause contain the most useful message.

By the way, i think tracing and logging should be considered separately. You can find the full exception stacktrace with traceId in log center.

emeroad:
I think logging and tracing should be tightly integrated.
It would be ideal to store the exceptions in a separate table and display the timeline.

IluckySi:
In the future, I will try to save the exception stack information in an HBase table separately ~ 
the perspective of log tool and APM tool in analyzing application performance problems is different.  They are complementary. The analysis of request exception is an important function for both log tool and APM tool.

PR #7199, opened 2020-09-01
Reduce memory usage of getAgentList

yjqg6666:
The from query parameter is not used to filter agent.

And the default getAgentList API for the agent statistics should use the from/to version for better performance. The current one will return too much data, in our case, 800MB data returned from /getAgentList.pinpoint and unable to be displayed.


emeroad:
@yjqg6666 Thank you for your contribution

This patch makes sense.
However, it seems that this API is not being used.
Can you tell me where you are using it? Do you use it internally?

yjqg6666:
We made some change, let https://github.com/naver/pinpoint/blob/master/web/src/main/java/com/navercorp/pinpoint/web/controller/AgentInfoController.java#L61-L62 to call getAgentList(from,to), default to some time period. It's would be better to update UI part, the user can select a time period until now or time range. 
http://125.209.240.10:10123/config/agentStatistic 
image

PR #7219, opened 2020-09-04
Fix typo and simplify method

dengliming:


emeroad:
@dengliming 

Issue #7258, opened 2020-09-11
@Override annotation is missing in pinpoint-web


imbf:
@Override annotation is missing in AgentChecker, DataSourceConnectionUsageRateChecker, DeadlockChecker

I think that when you override getSmsMessage method from abstract class you should add @Override annotation

imbf:
Issue close because related issues have been resolved

PR #7259, opened 2020-09-11
[#7258] Add @Override annotation in checker

imbf:
Add @Override annotation to getSmsMessage method in checker


emeroad:
@imbf 

Issue #7150, opened 2020-08-20
Add paho.mqtt.java plugin


acafela:
**Which module is your feature request related to?**
Plugin

**Is your feature request related to a problem?**
No

**Describe the solution you'd like**
Add [eclipse/paho.mqtt.java](https://github.com/eclipse/paho.mqtt.java) plugin for tracking [MQTT protocol](https://en.wikipedia.org/wiki/MQTT).

RoySRose:
@acafela 
any updates?

acafela:
@RoySRose 
I was working in the local.
I dug the library. Now it is possible to trace simple method in web.
I think it will take about 3 weeks to send the first PR, is that okay?

RoySRose:
@acafela 

> I think it will take about 3 weeks to send the first PR, is that okay?

Of course~! It's your time~ :) I was just curious~

PR #7266, opened 2020-09-15
[#7150] Add paho.mqtt plugin

acafela:
[eclipse/paho.mqtt.java](https://github.com/eclipse/paho.mqtt.java) plugin for trace mqtt publish, subscribe.

### modules
- plugins/paho-mqtt
- plugins-it/paho-mqtt-it
- agent-testweb/paho-mqtt-plugin-testweb

### type-provider
- code : 8501 / name : PAHO_MQTT_CLIENT
- code : 8502 / name : PAHO_MQTT_CLIENT_INTERNAL

### RPC Name
 mqttt://topic={topic}&qos={qos}

Screen Shot 2020-09-15 at 7 15 51 PM

### Call Stack
Screen Shot 2020-09-15 at 11 30 24 PM

### MQTT Version 3 Issue
MQTT Version 3 does not support custom properties,
***so it is impossible to map publication and subscription into a single transaction.***
[MQTT Version 3.1.1 specification](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718018)

### MQTT Version 5
MQTT Version 5 supports custom properties,
***so publication and subscription can be mapped into a single transaction.***


koo-taejin:
Hi @acafela 

First of all, I really thanks for your contribution. 

But, this PR build has failed on Travis. 
Would it be possible to check it. 

Thanks :)

acafela:
Thank you for your feedback. 
I'll check and update it!

koo-taejin:
@acafela 

This is plugin developer guide. 
https://naver.github.io/pinpoint/plugindevguide.html#222-continue-transaction

As you could see, basically Endpoint field has to set `Server(current node) address`.

In queue servicetype case, Pinpoint handles it as virtual node, so consumer is client, but it looks like as a server. 
In this part, if you include EndPoint as the server address of the queue, you could meet some problems.

If you look at the processing of rabbitmq, rabbitmq gets LocalAddress and hancles it.
https://github.com/naver/pinpoint/blob/master/plugins/rabbitmq/src/main/java/com/navercorp/pinpoint/plugin/rabbitmq/client/interceptor/RabbitMQConsumerDispatchInterceptor.java#L202

Kafka have been made like PR that you seneded.
I tried to get extract localaddress to use endPoint, but structure was too complex so I could not handle it right. 
So, Kafka Plugin has some problem because of it. 
I tried to changed it again in Kafka Plugin.

Please check it again, and If you could get local address then Please use it instead of server address of the queue at EndPoint. 

thanks :)

acafela:
@koo-taejin 
First of all, thank you for your detailed feedback and explanation 😀

1. EndPoint problem
I understand the problem you mentioned.
I will find a way to intercept the current node local address and update it.

2. License and author
I'll add and update it.

Thank you!

acafela:
@koo-taejin 

I fixed the wrong EndPoint problem.
The updated source accesses **_the socket inside the library_**.
Then get the host, port and record the endPoint.
**I referred to [https://github.com/naver/pinpoint/pull/7285](https://github.com/naver/pinpoint/pull/7285)**

And I added license, author to all Java files.

Thank you :)

plugins/paho-mqtt/src/main/java/com/navercorp/pinpoint/plugin/paho/mqtt/interceptor/CommsCallbackConstructorInterceptor.java
```diff
@@ -0,0 +1,91 @@
...
+
+    private String extractBrokerUri(Object[] args) {
+        // MQTT Version 3 case
+        if(this.config.isEnableTracePahoMqttClientV3() && args[0] instanceof org.eclipse.paho.client.mqttv3.internal.ClientComms){
+            org.eclipse.paho.client.mqttv3.internal.ClientComms clientComms = (org.eclipse.paho.client.mqttv3.internal.ClientComms)args[0];
+            org.eclipse.paho.client.mqttv3.IMqttAsyncClient mqttAsyncClient = clientComms.getClient();
+            return mqttAsyncClient.getServerURI();
+        }
+        // MQTT Version 5 case
+        if(this.config.isEnableTracePahoMqttClientV5() && args[0] instanceof org.eclipse.paho.mqttv5.client.internal.ClientComms){
```
koo-taejin: I could not find mqttv5 package in v1.0.0 ~ v1.2.0.
If user uses that versions, then it might occur ClassNotFoundException.

Even though code would be complicated, it is judged that the problem can be solved by registering the Interceptor class separately likes below.
(
"org.eclipse.paho.mqttv5.client.internal.ClientComms" -> CommsV3CallbackConstructorInterceptor
"org.eclipse.paho.mqttv3.client.internal.ClientComms" -> CommsV5CallbackConstructorInterceptor
)

acafela: @koo-taejin You're right. I missed the case.

I created two classes: _CommsCallbackV3ConstructorInterceptor and CommsCallbackV5ConstructorInterceptor_.

I was thinking of eliminating source duplication by making _CommsCallbackConceptorInterceptor_ an abstract class.
But I didn't do it because I thought it would make it more complicated.

koo-taejin:
@acafela 

Everything looks good. 

Please squash all commits that you have sent to one commit, then I am going to merge it. 

thanks :)

RoySRose:
Just a tip
It does not always have to be in one commit. 
Usually, commits are divided on a functional basis or sometimes a code basis. (For Pinpoint, I think it's forme)

However, since this is an initial commit for the plugin it would be fine as 1 commit as @koo-taejin said.

Also, please update the supported module document file for this contribution (https://github.com/naver/pinpoint/blob/master/doc/modules.md) and add `readme.md`, if there is anything you need to inform the user on the plugin

acafela:
@koo-taejin I will squash all commits including update the module document(@RoySRose mentioned).
Thank you :)

acafela:
@RoySRose Thanks for the tip!!
I forgot to update the document you told me.

I will also update this document and commit it.
I have created a [readme](https://github.com/naver/pinpoint/pull/7266/files#diff-7033f2f45513f8441b09acf4c50f5a88) for the module , so I'll link it to the module document.

acafela:
@koo-taejin I've squashed all commits. 
This commit also includes an update to `modules.md`.

koo-taejin (approved):
lgtm 👍

acafela:
> lgtm 👍

I am glad to hear that.
I hope I can contribute again. 🙏

PR #7271, opened 2020-09-17
[#noissue] Add SpringSmtpMailSender Unit Test

imbf:
case 1 : Constructor requireNonNull Test.
case 2 : When Sender Email Address is not configured, Constructor throw NullPointerException.
case 3 : When Sender Email Address is wrong, Constructor throw RuntimeException.
case 4 : When receiver does not exist, do not send email
case 5 : When receiver is wrong, do not send email
case 6 : send Email with AlarmChecker
case 7 : send Email with AgentChecker


minwoo-jung:
I will review soon.

imbf:
if you don't mind, may i ask for a review?

minwoo-jung:
@imbf
Sorry for the late confirmation. LGTM

imbf:
Thank you @minwoo-jung

PR #7274, opened 2020-09-18
collector, update local stat log filename format

yjqg6666:
Include month in log filename.

### DIFF
--- a/collector/src/main/resources/profiles/local/log4j2.xml
+++ b/collector/src/main/resources/profiles/local/log4j2.xml
@@ -38,7 +38,7 @@
         </RollingFile>
 
         <RollingFile name="stat-appender" filename="${logging_dir}/pinpoint-stat.log"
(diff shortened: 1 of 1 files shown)


collector/src/main/resources/profiles/local/log4j2.xml
```diff
@@ -38,7 +38,7 @@
         </RollingFile>
 
         <RollingFile name="stat-appender" filename="${logging_dir}/pinpoint-stat.log"
-                     filepattern="${logging_dir}/pinpoint-stat-%d{yyyy-dd-mm}-%i.log" >
+                     filepattern="${logging_dir}/pinpoint-stat-%d{yyyy-MM-dd-mm}-%i.log" >
```
emeroad: I think `${rolling-date-format}` would be better.

yjqg6666: Agree and just updated the PR. @emeroad

PR #7281, opened 2020-09-22
update sampler to support sampling rate range from 1% to 100%

yjqg6666:
1. The sampling rate now is n%, where n is an integer and ranges from 1 to 100.
2. If the sampling rate is 10%, 20% or some other rates which have no mod divided by 10, the requests would be checked every 10 requests(This would be very useful for testing). Otherwise they would be checked by every 100 requests.

Update Guide:
**Need to update agent config profiler.sampling.rate to use the new way.**

### DIFF
--- a/agent/src/main/resources/profiles/local/pinpoint.config
+++ b/agent/src/main/resources/profiles/local/pinpoint.config
@@ -56,8 +56,8 @@ profiler.jvm.stat.collect.detailed.metrics=true
 # Allow sampling.
 profiler.sampling.enable=true
 
-# 1 out of n transactions will be sampled where n is the rate. (1: 100%, 20: 5%)
-profiler.sampling.rate=1
+# n out of 100 transactions will be sampled. (1: 1%, 20: 20%, 100:100%)
+profiler.sampling.rate=100
 
 # Permits per second, if throughput is 0, it is unlimited.
 # "New" is a transaction that is newly traced.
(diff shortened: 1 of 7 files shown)


emeroad:
Does this sampler have a uniform distribution?

yjqg6666:
It's similar to the original one.

yjqg6666:
The sampler can be simplified by:
`
boolean shouldBeSampled = (count % 100) <= samplingRate
`

yjqg6666:
If the sampling rate is 50%, the first 5 requests of 10 requests would be traced using the new sampler.
The original sampler is that the odd request would be request.

emeroad:
The sampler should have as uniform distribution as possible. 
If the sampling is skewed, it will affect the distribution of the scatter chart.

I think it's better to think of a different algorithm.

yjqg6666:
> The sampler should have as uniform distribution as possible.
> If the sampling is skewed, it will affect the distribution of the scatter chart.
> 
> I think it's better to think of a different algorithm.

@emeroad 
Agreed. You remind me another solution, may have some performance penalty(have some complexities). We can reference Facebook's [PlanOut](https://facebook.github.io/planout/docs/random-operators.html) which have a UniformChoice random assignment operator. There is a port for Java language https://github.com/nkconnor/planout4j. I used this library for A/B testing client and it DO have uniform distribution.

yjqg6666:
I used weighted choice.
image

Here we need [uniform choice](https://github.com/nkconnor/planout4j/blob/master/core/src/main/java/com/glassdoor/planout4j/planout/ops/random/UniformChoice.java).

The same id will always have the same test result. It mainly [use sha1 hash]( https://github.com/nkconnor/planout4j/blob/master/core/src/main/java/com/glassdoor/planout4j/planout/ops/random/PlanOutOpRandom.java#L73).

PR #7415, opened 2020-11-13
[#noissue] PinpointPluginTestSuite JDK9+ compatibility

acafela:
During the plugin integration test, I found that PinpointPluginTestSuite doesn't work on jdk9+.

I modified this so that it can be executed on jdk9+.
Please let me know if it's Issues to be resolved directly by team or if it's not within the scope of pinpoint support.
Then I'll close it.

#### 1. In jdk 9+ environment, the `SharedProcessManager.fork()` running command does not have -cp argument.

This is because of the ClassLoader changed from jdk9+.
So i modified not to use URLClassLoader in jdk9+.

#### 2. When using @JvmVersion(9+) annontation

_class com.navercorp.pinpoint.bootstrapAgentBootLoader$1 cannot access class com.navercorp.pinpoint.test.PluginTestAgent_ exception occurred

This is because the _PluginTestAgent_ cannot be accessed from the external module. So I added it to the export list.

Thank you.

### DIFF
--- a/test/src/main/java/com/navercorp/pinpoint/test/plugin/AbstractPinpointPluginTestSuite.java
+++ b/test/src/main/java/com/navercorp/pinpoint/test/plugin/AbstractPinpointPluginTestSuite.java
@@ -20,6 +20,7 @@ import com.navercorp.pinpoint.common.Version;
 import com.navercorp.pinpoint.common.util.ArrayUtils;
 import com.navercorp.pinpoint.common.util.SystemProperty;
 import com.navercorp.pinpoint.exception.PinpointException;
+import com.navercorp.pinpoint.test.plugin.util.StringUtils;
 import org.eclipse.aether.resolution.ArtifactResolutionException;
 import org.eclipse.aether.resolution.DependencyResolutionException;
 import org.junit.internal.runners.statements.RunAfters;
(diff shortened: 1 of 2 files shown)


test/src/main/java/com/navercorp/pinpoint/test/plugin/AbstractPinpointPluginTestSuite.java
```diff
@@ -157,10 +162,33 @@ private String resolveTestClassLocation(Class<?> testClass) {
...
-        List<String> result = new ArrayList<String>();
-
         ClassLoader cl = getClass().getClassLoader();
+        if (cl.getClass().getName().startsWith(JAVA_9_INTERNAL_CLASSLOADER_PREFIX)) {
+            return getJava9ClassPathList(classPathCandidates);
+        }
+        return getClassPathList(cl, classPathCandidates);
+    }
+
+    private List<String> getJava9ClassPathList(String[] classPathCandidates) {
```
emeroad, line 172: The design of classloader dependent code seems to be bad.
If everything is fine, your code looks much better.
Could you replace the old code with yours?


```java
     cl = cl.getParent();
     if (cl == null) {
         break;
      }
```
There may be a problem with the above code, but I'm not sure.
Please check that it works properly in all tests.

acafela: Thank you for your review @emeroad
I'll change the old code and make sure there are no problems with all the tests.
When this test is completed i will commit again :)

smilu97:
resolved in #9416. Thank you for issuing!

PR #7462, opened 2020-12-01
Add C/C++ servertype

eeliu:
Add C/Cpp  icon


tankilo:
> Add C/Cpp icon

Hello, does this issue mean that C/Cpp application can be or planed to be traced?

Issue #7471, opened 2020-12-02
RabbitMq: Disconnect producer and consumer optionally


praveen12bnitt:
**Which module is your feature request related to?**
RabbitMq Plugin

**Is your feature request related to a problem?**
We run a highly distributed system with lot of messaging involved. So a single UI transaction will produce a lot of rabbitmq message and each of these messages in-turn will turn produce messages/call rest api etc.  

**Describe the solution you'd like**
Because pinpoint rabbitmq by default links the producer and consumer together, our transaction becomes very very large. Due to this, when we load a transaction is loaded in the UI it fails to load because of timeout. Think about a single transaction that goes via 10 different server and a lot of rest api/db calls made. 

**Describe alternatives you've considered**
One possible solution to this problem is to have a optional flag in the rabbitmq plugin that will not link the producer and consumer as a single transaction. 

So if app1 -> rabbitq1 -> app2 -> rabbit2 -> app3

This today will create a single transaction that will start in app1. When we break them, it will create
tran1 : app1-> rabbitq1
tran2 : rabbitq1-> app2 -> rabbit2
tran2 : rabbit2 -> app3

Since the transactions are split, the UI only loads a subset of api metadata/ sql metadata etc. 

**Additional context**

praveen12bnitt:
I created a PR for this. We tested this in our env and seeing positive results. Please review the proposal and let me know if you  need any changes.

praveen12bnitt:
@emeroad @koo-taejin can one of you review the PR and give me some feedback?

PR #7472, opened 2020-12-02
[#7471] add ability to disconnect rabbit producer and consumer

praveen12bnitt:
The new property
`profiler.rabbitmq.client.consumer.disconnect.enable`
which will determine if we need to continue the transaction at the consumer side.

So if app1 -> rabbitq1 -> app2 -> rabbit2 -> app3

When profiler.rabbitmq.client.consumer.disconnect.enable is true
we break them and create

tran1 : app1-> rabbitq1
tran2 : rabbitq1-> app2 -> rabbit2
tran3 : rabbit2 -> app3

### DIFF
--- a/agent/src/main/resources/profiles/local/pinpoint.config
+++ b/agent/src/main/resources/profiles/local/pinpoint.config
@@ -1068,6 +1068,8 @@ profiler.netty.channel.close=true
 profiler.rabbitmq.client.enable=true
 profiler.rabbitmq.client.producer.enable=true
 profiler.rabbitmq.client.consumer.enable=true
+# If you decided to break the producer - consumer chained tracing, set this to true.
+profiler.rabbitmq.client.consumer.disconnect.enable=false
 # Custom consumer classes to be traced (comma separated list of fully qualified class names)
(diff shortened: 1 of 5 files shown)


koo-taejin:
hi @praveen12bnitt 

First of all, thank you for your contribution.

I understand what you intentions.
But I don't know if it is best way to solve it. 
Nevertheless, I agree on some point for temporary solution until find out solution.

I will give you a response after thinking about the key name of configuration. 

thanks :)

praveen12bnitt:
If we can link the transactions together and still load the UI fast, then that will be the best solution. I agree with u. This is more of a temporary solution.

Please leave a comment here on the changes u want me to do and I will be more than happy to do it.

koo-taejin:
@praveen12bnitt 

How about setting the name to `profiler.rabbitmq.client.consumer.continue.enable` instead of `profiler.rabbitmq.client.consumer.disconnect.enable`?
I think it is better because Pinpoint use word of continue in many places, so users could have a good understanding of the settings
And the default setting is true, and true seems to work as before.
Please, do work I mentioned.

thanks :)

PR #7529, opened 2021-01-04
Add RocketMQ 4.x plugin.

messi-gao:


tankilo:
@messi-gao  does this plugin update support [Broadcasting](http://rocketmq.apache.org/docs/broadcast-example/) occasions ? 
Before, when a message is consumed by multi consumers, they will share the same spanId. I don't know how it is going on now.

messi-gao:
@tankilo It supports broadcasting occasions.I have tried the callstack is normal.

koo-taejin:
@messi-gao 
Thanks for your PR.
I am going to review this PR. :)

thanks :)

plugins/rocketmq/src/main/java/com/navercorp/pinpoint/plugin/rocketmq/interceptor/ProducerSendInterceptor.java
```diff
@@ -0,0 +1,306 @@
...
+            Trace trace = traceContext.currentTraceObject();
+            final Object sendCallback = getSendCallback(args);
+            // async send process
+            if (sendCallback instanceof AsyncContextAccessor) {
+                if (isSkipTrace()) {
+                    return;
+                }
+                final AsyncContext asyncContext =
+                        ((AsyncContextAccessor) sendCallback)._$PINPOINT$_getAsyncContext();
+                ((AsyncContextAccessor) target)._$PINPOINT$_setAsyncContext(asyncContext);
```
koo-taejin: I have been wondering where to use this.
I could not find place that used it.

koo-taejin:
@messi-gao 
From my analysis, it appears to be tracking only async type message.
If I analyzed correctly, I wonder if it is intended that you do not implement sync type message handling.

Please, answer. :)

plugins/rocketmq/src/main/java/com/navercorp/pinpoint/plugin/rocketmq/interceptor/ProducerSendInterceptor.java
```diff
@@ -0,0 +1,306 @@
...
+
+            for (Map.Entry<String, Object> entry : paramMap.entrySet()) {
+                properties.append(entry.getKey());
+                properties.append(NAME_VALUE_SEPARATOR);
+                properties.append(entry.getValue());
+                properties.append(PROPERTY_SEPARATOR);
+            }
+            sendMessageRequestHeader.setProperties(properties.toString());
+        } else {
+            paramMap.put(Header.HTTP_SAMPLED.toString(), SAMPLING_RATE_FALSE);
```
koo-taejin, line 194: When you don't sample, it looks like you don't include properties in the header.
please check.

plugins/rocketmq/src/main/java/com/navercorp/pinpoint/plugin/rocketmq/interceptor/ProducerSendInterceptor.java
messi-gao: You are right,I 'll remove it.

messi-gao: > I have been wondering where to use this.
> I could not find place that used it.

You are right,I 'll remove it.

plugins/rocketmq/src/main/java/com/navercorp/pinpoint/plugin/rocketmq/interceptor/ProducerSendInterceptor.java
messi-gao: > When you don't sample, it looks like you don't include properties in the header.
> please check.

Is there a need to pass properties when don't sample?

koo-taejin: Yes, it will be trigger that target server will be not sampled.

plugins/rocketmq/src/main/java/com/navercorp/pinpoint/plugin/rocketmq/interceptor/ProducerSendCallBackInterceptor.java
```diff
@@ -0,0 +1,106 @@
...
+                                       Object[] objects) {
+        }
+
+        @Override
+        protected void doInAfterTrace(SpanEventRecorder spanEventRecorder, Object target, Object[] args,
+                                      Object result, Throwable throwable) {
+            spanEventRecorder.recordServiceType(RocketMQConstants.ROCKETMQ_CLIENT_INTERNAL);
+            spanEventRecorder.recordApi(methodDescriptor);
+            final SendStatus sendStatus = ((SendResult) args[0]).getSendStatus();
+            if (throwable != null) {
```
koo-taejin, line 77: Like HttpStatusCode, how about using the AnnotationKey to record the SendStatus?
If an exception occurs when executing CallBack code, it is better if the user knows that callback has what kind of SendStatus occurred at the time the exception occurred.

In addition, if there is no exception in the callback, it seems okay to execute a recordException that is not SEND_OK as it is now.

messi-gao: > Like HttpStatusCode, how about using the AnnotationKey to record the SendStatus?
> If an exception occurs when executing CallBack code, it is better if the user knows that callback has what kind of SendStatus occurred at the time the exception occurred.
> 
> In addition, if there is no exception in the callback, it seems okay to execute a recordException that is not SEND_OK as it is now.

```
        recorder.recordApi(methodDescriptor);
        final ConsumeConcurrentlyStatus status = (ConsumeConcurrentlyStatus) result;
        recorder.recordAttribute(RocketMQConstants.ROCKETMQ_SEND_STATUS_ANNOTATION_KEY, status.name());
        if (throwable != null) {
            recorder.recordException(throwable);
        }
```
like this?

koo-taejin: Yes, It looks good. :) 

But, judging method with watching interface, result will be null. 
=>
`final ConsumeConcurrentlyStatus status = (ConsumeConcurrentlyStatus) result;`

messi-gao:
@tankilo I have fixed the comments above,please have a check.

messi-gao:
@tankilo I have fixed the comments above,please have a check.

plugins/rocketmq/src/main/java/com/navercorp/pinpoint/plugin/rocketmq/RocketMQPlugin.java
```diff
@@ -0,0 +1,200 @@
...
+    public static class MessageListenerConcurrentlyTransform implements TransformCallback {
+
+        @Override
+        public byte[] doInTransform(Instrumentor instrumentor, ClassLoader classLoader, String className,
+                                    Class<?> classBeingRedefined, ProtectionDomain protectionDomain,
+                                    byte[] classfileBuffer) throws InstrumentException {
+            final InstrumentClass target = instrumentor.getInstrumentClass(classLoader, className,
+                                                                           classfileBuffer);
+            final List<InstrumentMethod> consumeMessageMethods = target.getDeclaredMethods(
+                    MethodFilters.name("consumeMessage"));
```
koo-taejin: In my opinion, it would be better if you don't use MethodFilter like this.
In this case, it must execute the method declared in the interface.

In general, when there is a Boundary level scope, it is a case of many overloading.
This is because the user does not know which method to call.

Like below.
```
public void invoke() {
invoke("hello");
}

public void invoke(String message) {
invoke(message, null);
}

public void invoke(String message1, String message2) {
// todo
}
```

However, in this case, there is no need to do this because there are methods that must pass through the interface.

And, If you don't use it, you can reduce complex and you can reduce calculation time.

messi-gao: > In my opinion, it would be better if you don't use MethodFilter like this.
> In this case, it must execute the method declared in the interface.
> 
> In general, when there is a Boundary level scope, it is a case of many overloading.
> This is because the user does not know which method to call.
> 
> Like below.
> 
> ```
> public void invoke() {
> invoke("hello");
> }
> 
> public void invoke(String message) {
> invoke(message, null);
> }
> 
> public void invoke(String message1, String message2) {
> // todo
> }
> ```
> 
> However, in this case, there is no need to do this because there are methods that must pass through the interface.
> 
> And, If you don't use it, you can reduce complex and you can reduce calculation time.

Fix it.
BTW,when intercept an internal interface, it should use the `InterfaceInternalNameMatcherOperand `and must have a `basepackage`,like [this](https://github.com/messi-gao/pinpoint/blob/feature/rocketmq/plugins/rocketmq/src/main/java/com/navercorp/pinpoint/plugin/rocketmq/RocketMQPlugin.java#L75).is there a way to avoid pass the `basepackage `param?

plugins/rocketmq/src/main/java/com/navercorp/pinpoint/plugin/rocketmq/interceptor/ProducerSendCallBackInterceptor.java
koo-taejin: It looks good at having changed Consumer side. :)
Please check about Producer side in ProducerSendCallBackInterceptor.

koo-taejin: It looks good at having changed Consumer side. Please check about Producer side in ProducerSendCallBackInterceptor.

plugins/rocketmq/src/main/java/com/navercorp/pinpoint/plugin/rocketmq/interceptor/ProducerSendInterceptor.java
koo-taejin: I think it worked out well.
If it was deleted, then I think `target.addField(AsyncContextAccessor.class); in MQClientAPIImplTransform` looks to be not necessary. 
Please check it :)

plugins/rocketmq/src/main/java/com/navercorp/pinpoint/plugin/rocketmq/RocketMQPlugin.java
koo-taejin: I am not familiar with RocketMQ. 
So, I am going to answer it, after I research RockerMQ structure. :)

plugins/rocketmq/src/main/java/com/navercorp/pinpoint/plugin/rocketmq/interceptor/ProducerSendInterceptor.java
messi-gao: fix it.

plugins/rocketmq/src/main/java/com/navercorp/pinpoint/plugin/rocketmq/interceptor/ProducerSendCallBackInterceptor.java
messi-gao: > It looks good at having changed Consumer side. :)
> Please check about Producer side in ProducerSendCallBackInterceptor.

fix it.

koo-taejin:
@messi-gao 
As far as I have judged, one MessageExt means one request. 
But, RockerMQ seems that always bring it as a list, because of performance. 

It seems that you choose first message among them and make trace by it. 
```
        private static TraceFactory get(List<MessageExt> msgs) {
            final String traceId = msgs.get(0).getUserProperty(Header.HTTP_TRACE_ID.toString());
            if (StringUtils.isEmpty(traceId)) {
                return new DefaultTraceFactory();
            } else {
                return new SupportContinueTraceFactory();
            }
        }
```

In this case, Servermap and CallStack could show wrong data. 
(Ex :
```
	Client A -----> Message A                  ------------------↘
	Client B -----> Message B                  ------------------->  Consumer (A, B, C) 
	Client C -----> Message C                 -------------------↗
```	
	In this case, Pinpoint only shows  Client A -> Consumer (B and C could not express in ServerMap and Continue Trace.)
)

If I analyze correctly, I think that It would rather than such expression as Kafka Plugin than now. 

I am not familiar with RocketMQ, so it could possible that I have analyzed wrong. 
Please, feel free to say to your opinion or explanation.

messi-gao:
@tankilo Thanks for your reminder,I'll research about it.

messi-gao:
@tankilo 
Your analysis is correct.

The RocketMQ consumer will consume only one message by default,but it change the number by modifying the [consumeMessageBatchMaxSize](https://github.com/apache/rocketmq/blob/master/client/src/main/java/org/apache/rocketmq/client/consumer/DefaultMQPushConsumer.java#L221) property .

So I'll change the implemention like Kafka Plugin.

messi-gao:
@tankilo I have changed the trace of consumer,please check when you are have time.

plugins/rocketmq/src/main/java/com/navercorp/pinpoint/plugin/rocketmq/interceptor/ProducerSendCallBackInterceptor.java
koo-taejin: Please check it again. It seems that it did not be changed.

koo-taejin: Please check it again. It seems that it did not be changed.

messi-gao: I'am sorry.I fixed it just now.

koo-taejin:
@messi-gao 
For testing this plugin, Could you are going to add test environment at agent-testweb (https://github.com/pinpoint-apm/pinpoint/tree/master/agent-testweb)

messi-gao:
> @messi-gao
> For testing this plugin, Could you are going to add test environment at agent-testweb (https://github.com/pinpoint-apm/pinpoint/tree/master/agent-testweb)

Got it.I'll research about it and notice you once compelete it.

messi-gao:
@tankilo I have added some basic test about RocketMQ plugin.

messi-gao:
@tankilo I've added test environment at agent-testweb which includes original、springboot、springcloud stream using types.

koo-taejin:
@messi-gao 
Thanks you for this job. I am going to check this job. :)

koo-taejin:
@messi-gao 
I am happy with your PR, Most of the part.
But, please recheck a cople of things. 

1. Using Scope 
    I wonder if you must use Scope in Consumer. Please understand the role of Scope and check it again. But This thing will be up to your judgment.
2. Getting EndPoint
    For now, I checked to bring the EndPoint in this way. (`String endPointAddress = NetworkUtils.getHostIp();`)
    If you can possible, It needs more precise function. 
    Especially in the case of queues, they are more sensitive to this information because they use virtual nodes.
    So, I beg you to find out more good solution. 

3. Please change all copyright year to 2021.  

thanks :)

messi-gao:
> @messi-gao
> I am happy with your PR, Most of the part.
> But, please recheck a couple of things.
> 
> 1. Using Scope
>    I wonder if you must use Scope in Consumer. Please understand the role of Scope and check it again. But This thing will be up to your judgment.
> 2. Getting EndPoint
>    For now, I checked to bring the EndPoint in this way. (`String endPointAddress = NetworkUtils.getHostIp();`)
>    If you can possible, It needs more precise function.
>    Especially in the case of queues, they are more sensitive to this information because they use virtual nodes.
>    So, I beg you to find out more good solution.
> 3. Please change all copyright year to 2021.
> 
> thanks :)

1. Fixed.

2. It makes me confused when I get `EndPoint` field.It's not a good way to get it for now.But I still don't get a better way.I'll research about it.

3. Fixed.

plugins/rocketmq/src/main/java/com/navercorp/pinpoint/plugin/rocketmq/interceptor/ProducerSendCallBackInterceptor.java
```diff
@@ -0,0 +1,109 @@
...
+        @Override
+        protected void doInAfterTrace(SpanEventRecorder spanEventRecorder, Object target, Object[] args,
+                                      Object result, Throwable throwable) {
+            spanEventRecorder.recordServiceType(RocketMQConstants.ROCKETMQ_CLIENT_INTERNAL);
+            spanEventRecorder.recordApi(methodDescriptor);
+            final Object arg = args[0];
+            if (arg instanceof SendResult) {
+                final SendStatus sendStatus = ((SendResult) arg).getSendStatus();
+                spanEventRecorder.recordAttribute(RocketMQConstants.ROCKETMQ_SEND_STATUS_ANNOTATION_KEY,
+                                                  sendStatus);
```
koo-taejin: It is better to use sendStatus.name() instead of sendStatus.

koo-taejin:
> 2\. It makes me confused when I get `EndPoint` field.It's not a good way to get it for now.But I still don't get a better way.I'll research about it.

I am going to look at item 2 together. :)

thanks :)

plugins/rocketmq/src/main/java/com/navercorp/pinpoint/plugin/rocketmq/interceptor/ProducerSendCallBackInterceptor.java
messi-gao: fix it

koo-taejin:
@messi-gao 


I have checked that it is possible by using follow method. 

* Add getter at mQClientFactory field in DefaultMQPushConsumerImpl
```java

            transformTemplate.transform("org.apache.rocketmq.client.impl.consumer.DefaultMQPushConsumerImpl", ConsumerTransform.class);

            ->

            final InstrumentClass target = instrumentor.getInstrumentClass(classLoader, className, classfileBuffer);
            target.addGetter(MQClientInstanceGetter.class, "mQClientFactory");
```

* Add getter at channelTables field in NettyRemotingClient
```java
            transformTemplate.transform("org.apache.rocketmq.remoting.netty.NettyRemotingClient", RemotingTransform.class);

            ->

            final InstrumentClass target = instrumentor.getInstrumentClass(classLoader, className, classfileBuffer);
            target.addGetter(ChannelTablesGetter.class, "channelTables");
```

* location of NettyRemotingClient
```java
    mqClientInstance.getMQClientAPIImpl().getRemotingClient();
```

* result
```java
    ChannelTablesGetter._$PINPOINT$_getChannelTables().keySet();
    // {127.0.0.1:9876}
```

It is possible to extract by using map to attach messageListener.
(when execuing registerMessageListener() or start())

I'd like to know your opinion on this method.

thanks :)

messi-gao:
@tankilo 

I have a question about the `endPoint` field.As the [doc](https://github.com/pinpoint-apm/pinpoint/blob/master/doc/plugin-dev-guide.md#222-continue-transaction) describes,the `endPoint` field should be [Server(current node) address].

The result you get from here is the RocketMQ NameServer address,but not consumer address.
    ChannelTablesGetter._$PINPOINT$_getChannelTables().keySet();
    // {127.0.0.1:9876}

koo-taejin:
@messi-gao 
You're right.
Sorry, I think I misjudged it.
I'll check again to see if there's a good way.

thanks

koo-taejin:
@messi-gao 

I have checked that field of channelTables has NameServerInfo and BrokerServerInfo. 

If I add ChannelFutureGetter to ChannelWrapper. then I could get broker's channel.
```
            transformTemplate.transform("org.apache.rocketmq.remoting.netty.NettyRemotingClient$ChannelWrapper", ChannelWrapperTransform.class);
			->
			 target.addGetter(ChannelFutureGetter.class, "channelFuture");
```

And I could get NameServerList by `mqClientInstance.getClientConfig().getNamesrvAddr()`
If I remove namerserver in channelTables에서, then I could get EndPoint that Plugin needs.

But, this method is too complex, so I am not sure it is good....

messi-gao:
@tankilo 
The channelTables has NameServerInfo and BrokerServerInfo.you want to remove the NameServerInfo,so the BrokerServerInfo wil be left.I mean that we need **consumer address**,how can get the **consumer address** by the left BrokerServerInfo?

koo-taejin:
@messi-gao 
If I understand correctly, consumer address is local address connected to the broker.
(Producer -> Broker <- Consumer)

* channelTables 
```
(127.0.0.1:9876, 127.0.0.1:63917) -> nameserver
(10.25.215.103:10911, /10.25.215.103:65489) -> broker 
// 10.25.215.103:65489 <- it is local address connected to the broker
```
If it has multiple address, then {address1, address2, address3}

messi-gao:
@tankilo 
I know your meaning.

It's really a way to get the EndPoint.

But **nameserver** and **broker** could be multi.It may be complex to filter out the invalid data.

koo-taejin:
@messi-gao 
Good, then I think it doesn't have to have a filter. :)

messi-gao:
@tankilo 
I found that the `bornHost` filed of `MessageExt` may be the consumer address...

koo-taejin:
@messi-gao 
If you could access that way, it looks best solution. 
Because I am not familiar with RocketMQ, so I guess I've found a bad way.
Please proceed the way you found.

messi-gao:
@koo-taejin 
I'm really sorry that I always @ the wrong people.

I misjudged the `bornHost `filed of `MessageExt`,it's the **producer** address, not the **consumer** address.

I found that the RocketMQ use [RemotingUtil ](https://github.com/apache/rocketmq/blob/master/remoting/src/main/java/org/apache/rocketmq/remoting/common/RemotingUtil.java#L93)to get local address which can get IPV4 or IPV6 address.

Is it possible to record the `EndPoint ` that just contains IP(not contains port)?

I'll keep researching about it.

koo-taejin:
@messi-gao 
If possible, port should be included.

hays2hong:
@messi-gao How to deal with the transfer of trace in batch sending, because there is a bug in the RocketMQ versions below 4.8.0. https://github.com/apache/rocketmq/pull/2343

messi-gao:
> @messi-gao How to deal with the transfer of trace in batch sending, because there is a bug in the RocketMQ versions below 4.8.0. [apache/rocketmq#2343](https://github.com/apache/rocketmq/pull/2343)

Thanks for your reminding.

I found this bug during testing.

It's the broker server bug,the Pinpoint plugin can only intercept the client jar,so we may have no idea about it.

For RocketMQ versions below 4.8.0,the Pinpont plugin can only record the consumer trace and lose the Producer trace.

hays2hong:
@messi-gao Can we intercept  org.apache.rocketmq.client.producer.DefaultMQProducer#batch to resolve the problem?

messi-gao:
> @messi-gao Can we intercept org.apache.rocketmq.client.producer.DefaultMQProducer#batch to resolve the problem?

@hays2hong

We have intercepted the Producer and add Pinpoint trace info to Producer header properties.the Pinpoint trace info will send to broker server,the broker server send the trace info to Consumer.

If the broker could not send trace info to Cousumer,then the Consumer couldn't get trace info.

hays2hong:
@messi-gao 
We can add Pinpoint trace info to Message in org.apache.rocketmq.client.producer.DefaultMQProducer#batch. Then the properties will be serialized to byte[] and setted to MessageBatch#body. 
The Pinpoint trace will encoded in broker , and the key method is org.apache.rocketmq.store.CommitLog.MessageExtBatchEncoder#encode
image

messi-gao:
@hays2hong 

I don't think it's good way to put Pinpoint trace info into body.

we shoudn't change message body,it may occur unexpected error.

BTW,it's RocketMQ bug,we should not modify interceptor type to avoid this bug.You can upgrate the Rocketmq to solve it.

hays2hong:
@messi-gao 

You are right.

But it costs a lot to upgrate the Rocketmq in a company.

messi-gao:
@koo-taejin 
I have got the Consumer address by your opinion.It's really complex to get it.

Please have a check,thanks!

koo-taejin:
@messi-gao 
I appreciate your effort for this PR. 
If you do squash merge about all commit you posted, then I am going to merge it.
Thanks :)

messi-gao:
@tankilo I have done squash merge,please have a check. 
Thanks :)

koo-taejin:
@messi-gao 
Please remove this fc92403f33454a3fac586262f27ca208b512cb4d commit. 
We try to avoid merge commit. 

thanks :)

messi-gao:
> @messi-gao
> Please remove this [fc92403](https://github.com/pinpoint-apm/pinpoint/commit/fc92403f33454a3fac586262f27ca208b512cb4d) commit.
> We try to avoid merge commit.
> 
> thanks :)

@tankilo Fix it.

Thanks :)

messi-gao:
@koo-taejin 
I have found that the RocketMQ plugin code has been merged to the v2.3.0,could you add some description about it on release note?
Thanks.

koo-taejin:
@messi-gao 
Sorry for the slow reply. I'll check it out and add it to the release note. 

Thanks :)

koo-taejin:
@messi-gao 

I confirmed that I have missed the introduction about RocketMQ on 2.3.0 release notes.
If you post here images that related RocketMQ for using in the release notes, I am going to use them to add them to the release notes.

PR #7533, opened 2021-01-05
Agent adds configuration parameters to support environment variables

azi03:
Agent adds environment variable support of configuration file. It can set the configuration parameters needed by agent by setting system environment variable. System attribute (- D) has the highest priority, environment variable has the lowest priority, and configuration file has the lowest priority


emeroad:
Could you change Tab to 4 space?

azi03:
> Could you change Tab to 4 space?

OK, it's done

emeroad:
Please change the title to English.

azi03:
Supplementary note: the key of the environment variable is to replace the points in the configuration parameter with underscores, and all of them are capitalized, for example: profiler.transport.grpc.collector.ip , should be configured as PROFILER_TRANSPORT_GRPC_COLLECTOR_IP in the environment variable

Issue #7211, opened 2020-09-03
[Feature request] side-bar-container alongside main server-map can be folded upwardly/rightwardly


yjqg6666:
**Which module is your feature request related to?**
UI

**Is your feature request related to a problem?**
If the side-bar area can be hidden, the the whole main server map area could be larger for a better view, esp. when more than one inbound/outbounds are chosen.

yjqg6666:
![image](https://user-images.githubusercontent.com/1879641/92071483-6409a680-ede1-11ea-9198-51fe89874016.png)

binDongKim:
I agree. Maybe we could add a button at the middle-top of the left edge of the side-bar which will spread and fold the side-bar depending on the event.

yjqg6666:
@binDongKim I made a PR #7582 for it. PTAL.

PR #7582, opened 2021-01-26
[UI] The sidebar can be folded.

yjqg6666:
Related issue #7211 

image

When folded:
image

### DIFF
--- a/web/src/main/angular/src/app/core/components/side-bar/side-bar-container.component.css
+++ b/web/src/main/angular/src/app/core/components/side-bar/side-bar-container.component.css
@@ -2,6 +2,10 @@
     position: relative;
     width: 477px;
 }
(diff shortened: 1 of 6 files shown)


binDongKim:
I've tested out your codes and there are a couple of issues.

1. It would be nice if the server-map gets wider and smaller with the sidebar action.
2. The response-summary chart and the load chart renders weird when spreading back the sidebar after changing something(period, application etc)
3. I'm not sure about the UX itself at the moment. The button you added seems intuitive but it doesn't seem to go nicely with other elements(in my opinion).

So before going any further, I'd like to come up some ideas with you about the design first. I'll keep sharing/adding references(other examples) in this thread. If you find anything good, Please share it here. Thank you.

PR #7649, opened 2021-02-09
2.2.x

zhangyinhao1234:
添加了一个sdk，用来创建线程池，和创建runable，参考Spring的Async的实现方式，使用自己创建的runable和线程池实现了多线程链路的显示，原来这块在代码在1.8.3版本上运行，已经稳定运行1年。

### DIFF
--- a/pinpoint-java-sdk/pom.xml
+++ b/pinpoint-java-sdk/pom.xml
@@ -1,0 +1,76 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>pinpoint</artifactId>
+        <groupId>com.navercorp.pinpoint</groupId>
+        <version>2.2.2-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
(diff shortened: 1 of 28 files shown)


zhangyinhao1234:
example：

    private ThreadPoolExecutor triggerPool = new AsyncThreadPoolExecutor(
            32,
            256,
            60L,
            TimeUnit.SECONDS,
            new LinkedBlockingQueue<Runnable>(1000));

    
    @GetMapping("/hello1")
    public String test11(String name) throws InterruptedException {
        String hello = helloService.hello();
        triggerPool.execute(new AsyncRun() {
            @Override
            public Object asyncRun() {
                String pp = helloService.hello2(name, "PP");
                return pp;
            }
        }.getRunnable());
        return hello;
    }

zhangyinhao1234:
日志中未发现提示错误的地方。。不清楚哪里出了错

emeroad:
Please use English.

If there is a lot of log output from travis, the build will fail.

Could you push PR to master branch.
The master branch uses github action, so there is no such problem.

zlp1992:
@zhangyinhao1234 @emeroad  what's the difference between `pinpoint-java-sdk` and the existed `thread `plugin(since Pinpoint 2.0.2) ? I think the purpose of both is same, just to keep tracing in aysnc thread. Using `thread `plugin we don't have to modify business application, while using `pinpoint-java-sdk` we have to modify our business application to support async trace  (for example, use `AsyncRun` in our code) ?

emeroad:
The thread-plugin operates on global state, making it difficult to limit specific targets.
On the other hand, SDK can also have high accuracy.

For example, SDK makes it very easy to select a specific action when there are multiple asynchronous actions in a method.

zlp1992:
Get it. But using SDK we will have to modify our application code ?

Issue #7666, opened 2021-02-17
Compatibility with newer versions of Oracle JDBC driver


davide-parini:
I'm working with Pinpoint 2.0.4, OpenJDK 11, JBoss EAP 7.3.

JDBC Driver: Oracle Database 19c (19.8) ([link](https://download.oracle.com/otn-pub/otn_software/jdbc/198/ojdbc10.jar))

Oracle database trace entries are shown as "UNKNOWN_DB" while using the driver version mentioned above, while the older version of the Oracle JDBC Driver i was using before (11.2.0.3.0) works as expected.

After some debugging, i found the problem is with a new method in `oracle.jdbc.driver.OracleDriver`:
![image](https://user-images.githubusercontent.com/38081480/108209660-d8839800-712a-11eb-8bbc-b355ea23c91a.png)

The first `connect` method is the one that gets instrumented by Pinpoint Oracle JDBC Plugin, the only method present in older versions of the JDBC Driver. The second one is ignored. It seems that in newer versions of the driver the first `connect` method is not used (at least not in XA mode), so the plugin misses its execution.

What worked for me, for both driver versions, was something as simple (and ugly) as the following modification to `com.navercorp.pinpoint.plugin.jdbc.oracle.OraclePlugin$OracleDriverTransformer.doInTransform`. From:
```java
    public static class OracleDriverTransformer implements TransformCallback {
        @Override
        public byte[] doInTransform(Instrumentor instrumentor, ClassLoader loader, String className, Class<?> classBeingRedefined, ProtectionDomain protectionDomain, byte[] classfileBuffer) throws InstrumentException {
            InstrumentClass target = instrumentor.getInstrumentClass(loader, className, classfileBuffer);

            InstrumentUtils.findMethod(target, "connect",  "java.lang.String", "java.util.Properties")
                    .addScopedInterceptor(DriverConnectInterceptorV2.class, va(OracleConstants.ORACLE), ORACLE_SCOPE, ExecutionPolicy.ALWAYS);

            return target.toBytecode();
        }
    };
```
To:
```java
    public static class OracleDriverTransformer implements TransformCallback {

        private final PLogger logger = PLoggerFactory.getLogger(this.getClass());
        
        @Override
        public byte[] doInTransform(Instrumentor instrumentor, ClassLoader loader, String className, Class<?> classBeingRedefined, ProtectionDomain protectionDomain, byte[] classfileBuffer) throws InstrumentException {
[...]

PR #7683, opened 2021-02-19
[#7666] New connect method is now instrumented if present.

davide-parini:
Fixing as described in #7666.


ga-ram (approved):
Added connect method is present only in some versions of oracle database 19.
I tested with older oracle database versions with ojdbc6 and ojdbc8 in java8. 
The changes in this PR is compatible with older versions.

PR #7736, opened 2021-03-15

IluckySi:
I pulled a request that supports the collection of exception stack details, but for some reason, it didn't merge to the master (I guess the reason may be that I forked the pinpoint to an Organization, but I don't know how to set the allow edits from maintainers permission). Recently, I took the time to sort out this logic and put it directly under my user~

https://github.com/pinpoint-apm/pinpoint/pull/7255

image

### DIFF
--- a/agent/src/main/resources/pinpoint-root.config
+++ b/agent/src/main/resources/pinpoint-root.config
@@ -253,6 +253,11 @@ profiler.callstack.max.depth=64
 # weather or not to propagate exceptions occurred at interceptor
 profiler.interceptor.exception.propagate=false
 
+## Trace Exception StackTrace
+# Set line, default is enable and trace 20 line, if -1 is unlimited.
+profiler.exception.stacktrace.enable=true
+profiler.exception.stacktrace.line=20
+
 # Allow bytecode framework (ASM only)
 profiler.instrument.engine=ASM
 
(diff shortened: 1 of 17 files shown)


IluckySi:
It's really caused by this problem, that is, if fork to a certain organization, you can't see the "allow edits by maintainers" as shown in the figure~

image

IluckySi:
Pinpoint is a very good APM platform, it can help us quickly locate many potential performance problems of application, such as operating MySQL exception, accessing other services timeout and so on. During one use, I found that there was no detailed stacktrace information for the exceptions collected by the agent. In the issue, it was also mentioned that: #6902
Sometimes the detailed exception stack is also very important for us to quickly locate the problem, so I added a featuer, which includes the following two configuration items:
profiler.exception.stacktrace .enable=true # default is true
profiler.exception.stacktrace .line=20 # default is 20

Judge whether the function 'profiler.exception.stacktrace.enable' is enabled. If not, the detailed exception stack will not be collected; If enabled, collect by 'profiler.exception.stacktrace.line'. For example，the detailed test process is as follows:
profiler.exception.stacktrace.enable=false
profiler.exception.stacktrace.line=20
image
profiler.exception.stacktrace.enable=true
profiler.exception.stacktrace.line=5
image

IluckySi:
@emeroad  Are there any questions with this pr，thx~

emeroad:
I think the stack trace should be changed to a reference id like API-METADATA
```
Trace-Table
 \-span 
     |- exception message
```

Current PR
```
Trace-Table
 \-span 
     |- exception message + stack trace
```
This design has the problem of making the span table too large.


```
Trace-Table
 \- span 
     |- exception message
     |- stack traceId ref-id    -----------------------------------------
                                                                        |
Stack trace Table                                                       |
\- row-key : agentId + agentStartTime + stacktrace-id   <----------------
|-- column :  stacktrace 
```
What do you think of this design?

IluckySi:
It's a very good design, it will control the size of the whole trace; But I hope that as an optimization point, I  adjust it in the next pr.

Issue #7654, opened 2021-02-09
Pinpoint SDK for asynchronous operation


emeroad:
https://github.com/pinpoint-apm/pinpoint/pull/7649
@zhangyinhao1234 
Thank you for your contribution.

This approach is simple and very effective.

## ​​#7649 design
1. SDK provides Runnable and Executor classes
  - Now the target class to instrument is determined by the SDK
2. Async plug-in instrument the SDK class.

## ~~My design~~
1. Agent provides API for async operation.
2. SDK uses the agent's API to measure performance.

**Weakness**
- API should be provided through Boostrap-classloader. Its development is very complicated.
- Difficulty maintaining compatibility between SDK and agent API.

emeroad:
@zhangyinhao1234
Your PR disappeared as I pushed my rebase revision to the master branch.
It's my mistake. I have not found the cause of this problem. I'm really sorry.

I made a new branch and restored it.

zhangyinhao1234:
I'm glad my design was adopted

PR #7750, opened 2021-03-19
[#7654] Async support SDK

emeroad:


yjqg6666:
How about renaming the package name for the java SDK?

emeroad:
Maybe I will do it like below.
```java
import com.navercorp.pinpoint.sdk.v1.concurrent.xxx;
```

I plan to use the delegate pattern as much as possible.
```java
    public void test1() {

        ExecutorService executorService = TraceExecutors.wrapExecutorService(Executors.newSingleThreadExecutor());

        executorService.execute(TraceRunnable.wrap(new Runnable() {
            @Override
            public void run() {
                // action
            }
        }));
    }
```

However, ForkJoin class has a problem that cannot be delegation
```java
    public class ForkJoinPool extends AbstractExecutorService {
              public <T> ForkJoinTask<T> submit(Callable<T> task) {
              }
    }
```

There is also a way to use Annotation. This does not require any SDK dependency.

```java
// If the annotation class name is configurable, no SDK dependencies are required.
profiler.trace.asynctrace.classname=x.y.z.AsyncTrace

@AsyncTrace
public class MyRunnable implement Runnable {
     public void run() {
         // action
     }
}
```

Issue #7381, opened 2020-11-05
Auto generate AgentId


emeroad:
AgentID is no longer a requirement.

If AgentID is not present, the agent automatically generates it.

Agent log
```
11-05 17:22:13.251 INFO  AgentIdResolver                     : Failed to resolve AgentId(-Dpinpoint.agentId)
11-05 17:22:13.409 INFO  AgentIdResolver                     : 'BYDnVJbPR6ODxGK61qTT1A' AgentId is automatically generated
11-05 17:22:13.409 INFO  IdValidator                         : check SystemProperties(-D) applicationName:emeroad-app
11-05 17:22:13.410 INFO  AgentIdResolver                     : SystemProperties(-D) emeroad-app=emeroad-app
```
![image](https://user-images.githubusercontent.com/7564547/98226904-01c45200-1f9a-11eb-85f3-1647bab6057b.png)
![image](https://user-images.githubusercontent.com/7564547/98227024-1f91b700-1f9a-11eb-8ece-610bbf922f4c.png)

emeroad:
Due to this change, agentId has a problem that is not human friendly.
To solve this problem, we will add a human friendly agentName field.

## agentName 
- Not unique
   - AgentName is a view-only field and is not used as a primary key
- Size limit: 254 or 255
- display pattern : agentName (agentId)
   -Dpinpoint.agentName=MyBlogService.front.001 -Dpinpoint.agentId=BYDnVJbPR6ODxGK61qTT1A
     MyBlogService.front.001(BYDnVJbPR6ODxGK61qTT1A)

yjqg6666:
How about adding a configuration agentId.pattern=%hostname%-%ip%  (support %hostname%, %ip%, %uuid%). The end user can decide.

emeroad:
It's good. I have the same idea.

```
-DagentName=${hostname}-${ip}-${uuid}
```

The bracket(${...}) should have been shell friendly.
I don't know which one is good.

yjqg6666:
IMHO, configuration by config file is preferred. We have to configure JVM if the property is used.

yjqg6666:
Usually the JVM configuration is configured by the Operator/Sysadmin and the agent configuration can be configured by the developers themselves.

yjqg6666:
It would be also great that the applicationName and agentId can be configured by environment variable. 
[...]

PR #7752, opened 2021-03-22
Agent name support

yjqg6666:
- [x]   Agent part, Read -Dpinpoint.agentName / command argument --agentName / environment variable PINPOINT_AGENT_NAME as agent name. Record agent name as a field of agent information.

- [x]  Collector part, save agent name in agent information. 

- [x]  Web part, Show agent name in pinpoint web pages. 

Related to #7381 

The changes to UI are illustrated as follows:

image

image

image

image


commons-server/src/main/java/com/navercorp/pinpoint/common/server/bo/serializer/trace/v2/SpanDecoderV0.java
```diff
@@ -180,6 +180,9 @@ public void readSpanValue(Buffer buffer, SpanBo span, SpanDecodingContext decodi
 
         List<SpanEventBo> spanEventBoList = readSpanEvent(buffer, decodingContext, SEQUENCE_SPAN_EVENT_FILTER);
         span.addSpanEventBoList(spanEventBoList);
+        if (bitFiled.isSetAgentName()) {
+            span.setAgentName(buffer.readPrefixedString());
+        }
```
yjqg6666: @emeroad Here i have two solutions. I don't know which one is prefered.
Solution 1: as the committed. Occupied the last reserved bit of SpanBitFiled to indicate whether the agent name is set or not. Just write or read when the last bit is set. In this case the SpanBitField is used up. It's impossible for future features.

Solution 2: For the read side, just check for buffer.hasRemaining(), if true then readPrefixedString for agent name, the same as https://github.com/pinpoint-apm/pinpoint/pull/7752/files/9df60df694d373415404d69371bc8f3b1c897932..8b843872bb4dd550619544791fd6bad31cf3dc20#diff-355855262013205cf58c6c3dd88f4591dd8d002c8b52461ae6d1ff6dff15f5e8L78-L85 . 
For the write side, just putPrefixedString. 
Is the "remaining" solution prefered for compatibility?

Which one would you prefer? Solution 2 ? TIA.

yjqg6666:
Another question is:
It's impossible for the span chunks to save agent as in the current code base: 
https://github.com/pinpoint-apm/pinpoint/blob/3ab4595e3217b9010732c9f8ea01d3a35c5aea87/commons-server/src/main/java/com/navercorp/pinpoint/common/server/bo/serializer/trace/v2/SpanEncoderV0.java#L92-L108

yjqg6666:
Ignore the previous one, got the solution. The "remaining" solution.

commons-server/src/main/java/com/navercorp/pinpoint/common/server/bo/serializer/trace/v2/SpanDecoderV0.java
emeroad: In my opinion, there is a way not to store the agentName in the span.

How about including the  agentName in the agentInfo?
The agentName is stored in the agentInfo-table and can be queried using the agentId.

yjqg6666: Agent names are stored in both tables(agentInfo and trace tables).  I am thinking show agent name in the call tree. I will check if it's possible not to save agent name in span data and just query the agent info via the agentId just from the web part. It would be so much simpler and less changes to agent and collector.

yjqg6666:
@emeroad @binDongKim The feature is completed. Please take some time to review.

@emeroad In the PR i renamed the  AGENT_NAME_MAX_LEN to AGENT_ID_MAX_LEN in com.navercorp.pinpoint.common.PinpointConstants by refactoring rename to avoid confusion.

yjqg6666:
I will recheck for backwards-compatibility since some changes were made.

yjqg6666:
Recheck done.

yjqg6666:
I would like to make a New PR for a cleaner commit history for review and close this one.

PR #7766, opened 2021-03-31
Fix agent id resolver log

yjqg6666:
No feature involved.


emeroad:
@yjqg6666
Thanks for your contribution.

Issue #7795, opened 2021-04-12
[performance] reduce applications.pinpoint request from web UI


yjqg6666:
**Which module is your feature request related to?**
UI

**Is your feature request related to a problem?**
In almost every pinpoint page, it will (almost) always request the applications.pinpoint api event the the result may not be used, esp. in the tx list page and the tx mixed view page. Is there any reason to do this? If omitted, some hbase overhead may be reduced. 
**Describe the solution you'd like**
It a page do no need all the application list data, do NOT make the request.

binDongKim:
This is done by https://github.com/pinpoint-apm/pinpoint/pull/8059

PR #7796, opened 2021-04-12
[#7795] reduce unnecessary applications.pinpoint request from web UI

yjqg6666:
Remove unnecessary applications.pinpoint request on the following pages:
1. transaction list page
2. transaction view page
3. transaction detail page
4. filtered map page
5. browser not support page
6. scatter full screen mode page

The first four pages are the most-visited pages. Opt the request out should boot the page loading.

### DIFF
--- a/web/src/main/angular/src/app/app.routing.ts
+++ b/web/src/main/angular/src/app/app.routing.ts
@@ -17,16 +17,21 @@ const appRoutes: Routes = [
             {
                 path: '',
                 resolve: {
-                    configuration: SystemConfigurationResolverService,
-                    applicationList: ApplicationListResolverService
+                    configuration: SystemConfigurationResolverService
                 },
                 children: [
                     {
                         path: UrlPath.URL_STATISTIC,
+                        resolve: {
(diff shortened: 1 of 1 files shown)


yjqg6666:
@binDongKim PTAL.

binDongKim:
Your suggestion is reasonable so I've tried your approach but there was one thing I need to resolve.

We navigate to other pages with a new tab in most cases except the "config page". In this case, if I put the application resolver in the child route like you did, the application api call would keep requested when navigating to the config page and the main page back and forth. 

So I will think about the solution for the above issue first and then when it gets resolved, I will take your approach.
Many thanks.

yjqg6666:
@binDongKim The config page will still request the api, since in some child pages the data is still needed. The first 4 pages are the most-wanted, so i kept the config pages untouched. If needed, the resolver for config children pages should be enumerated.

binDongKim:
Yeah I know. But I mean, users go to the config page and back to the main page back and forth and the api would be request repeatedly which causes the time longer. But I get your point.

yjqg6666:
Now i get your point.

Issue #7803, opened 2021-04-15
incorrect link on application cell in agent statistics admin page


yjqg6666:
#### What version of pinpoint are you using?
master

#### Describe the bug
In the administration area, the agent statistics page, the application cell linked to /main/@STAND_ALONE due to the missing application name.
![image](https://user-images.githubusercontent.com/1879641/114837717-9747f580-9e06-11eb-9d02-7adc74ecb5c8.png)

PR #7805, opened 2021-04-15
[#7803] Add agent statistics table cell link

yjqg6666:
Changes:
1. always show application name in children rows and link to /main/{app_name}@{service_type}

before:

image

after:

image

2. the other cells except application cells link to /inspector/ {app_name}@{service_type} /5m/ {start_time + 5m} / {agent_id}

### DIFF
--- a/web/src/main/angular/src/app/core/components/agent-statistic-list/agent-statistic-list.component.ts
+++ b/web/src/main/angular/src/app/core/components/agent-statistic-list/agent-statistic-list.component.ts
@@ -7,6 +7,7 @@ export interface IGridData {
     serviceType: string;
     agent: string;
     agentVersion: string;
+    startTimestamp: number;
     jvmVersion: string;
     folder?: boolean;
     open?: boolean;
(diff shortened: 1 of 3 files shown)


web/src/main/angular/src/app/core/components/agent-statistic-list/agent-statistic-list.component.ts
```diff
@@ -88,7 +90,10 @@ export class AgentStatisticListComponent implements OnInit  {
                 field: 'agent',
                 width: 300,
                 filter: 'agTextColumnFilter',
-                tooltipField: 'agent'
+                tooltipField: 'agent',
+                cellStyle: {
+                    'cursor': 'pointer'
+                }
```
binDongKim: The changes you made work well! I think it would be better to apply the same cell-style on the agent field so that users can assume the agent field is also clickable.
```suggestion
                cellStyle: {
                    color: 'rgb(54, 162, 235)',
                    'font-weight': 600,
                    'cursor': 'pointer'
                }
```
ex)
스크린샷 2021-04-21 오전 11 24 02

binDongKim (changes requested):
I've left a comment. Thanks for your contribution.

PR #7823, opened 2021-04-23
[#noissue] spring plugin, trace async task executor submitListenable

yjqg6666:
Trace asyncTaskExecutor.submitListenable(Callable).


emeroad:

Issue #7825, opened 2021-04-23
spring async task traced can be configured


yjqg6666:
**Which module is your feature request related to?**
spring plugin

**Is your feature request related to a problem?**
The async task traced is currently on interceptor org.springframework.aop.interceptor.AsyncExecutionInterceptor and can not be configured.

**Describe the solution you'd like**
A PR #7826 is made.

PR #7826, opened 2021-04-23
[#7825] spring plugin, async task to be traced can be configured

yjqg6666:
see #7825


emeroad:

PR #7831, opened 2021-04-26
[#7803] update cell style in statistics table

yjqg6666:
add cell link style to agent name column & small refactoring.

### DIFF
--- a/web/src/main/angular/src/app/core/components/agent-statistic-list/agent-statistic-list.component.ts
+++ b/web/src/main/angular/src/app/core/components/agent-statistic-list/agent-statistic-list.component.ts
@@ -78,12 +78,8 @@ export class AgentStatisticListComponent implements OnInit  {
                     },
                     suppressCount: true
                 },
-                cellStyle: {
-                    color: 'rgb(54, 162, 235)',
-                    'font-weight': 600,
-                    'cursor': 'pointer'
-                },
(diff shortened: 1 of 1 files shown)


yjqg6666:
@binDongKim PTAL.

binDongKim:
I've tested it and it's displayed like below:
스크린샷 2021-04-27 오전 10 31 22

There seems to be 2 issues:

1. You applied the link-style and click event to the `Agent Name` field. What I'm curious is that the new tab which is opened through the `Agent Name` field uses the same url with the `Agent Id` field. Is this your intention?

2. I think you need to modify the cell-click logic part because now, if you click any field except the `Application` field, the new tab is always opened with the `Agent Id`. So I suggest you specify the clicked cell field here other than just `else`:
스크린샷 2021-04-27 오전 10 53 24

yjqg6666:
> What I'm curious is that the new tab which is opened through the Agent Name field uses the same url with the Agent Id field. Is this your intention?
Yes. It's my intention.
> So I suggest you specify the clicked cell field here other than just.
Other than the application field, the others are both link to the agent inspector page. If possible, i would like to make the agent version and agent jvm version cell clickable to make additional filter, but i haven't figure out how. Before that i intend to  make the others field link to the same.

binDongKim:
I got your point. Thank you.

PR #7854, opened 2021-05-10
[#noissue] Polishing

emeroad:


plugins/reactor-netty/src/main/java/com/navercorp/pinpoint/plugin/reactor/netty/interceptor/HttpClientOperationsOnInboundNextInterceptor.java
```diff
@@ -81,7 +82,7 @@ public void doInAfterTrace(SpanEventRecorder recorder, Object target, Object[] a
     }
 
     private boolean validate(final Object[] args) {
-        if (args == null || args.length < 2) {
+        if (ArrayUtils.getLength(args) != 1) {
```
koo-taejin, line 85: I think `args.length < 2` would be better to change  `ArrayUtils.getLength(args) < 2`.

plugins/reactor-netty/src/main/java/com/navercorp/pinpoint/plugin/reactor/netty/interceptor/HttpServerHandleHttpServerStateInterceptor.java
```diff
@@ -34,7 +35,7 @@ public HttpServerHandleHttpServerStateInterceptor(TraceContext traceContext, Met
     }
 
     public boolean validate(Object[] args) {
-        if (args == null || args.length < 2) {
+        if (ArrayUtils.getLength(args) != 1) {
```
koo-taejin, line 38: I think `args.length < 2` would be better to change `ArrayUtils.getLength(args) < 2`, args could be 0 size arrays.

plugins/reactor-netty/src/main/java/com/navercorp/pinpoint/plugin/reactor/netty/interceptor/HttpServerHandleStateInterceptor.java
```diff
@@ -33,7 +34,7 @@ public HttpServerHandleStateInterceptor(TraceContext traceContext, MethodDescrip
     }
 
     public boolean validate(Object[] args) {
-        if (args == null || args.length < 2) {
+        if (ArrayUtils.getLength(args) != 1) {
```
koo-taejin, line 37: I think `args.length < 2` would be better to change `ArrayUtils.getLength(args) < 2`, args could be 0 size arrays.

plugins/tomcat/src/main/java/com/navercorp/pinpoint/plugin/tomcat/interceptor/RequestStartAsyncInterceptor.java
```diff
@@ -72,8 +74,9 @@ public void after(Object target, Object[] args, Object result, Throwable throwab
             final SpanEventRecorder recorder = trace.currentSpanEventRecorder();
             if (validate(target, result, throwable)) {
                 com.navercorp.pinpoint.bootstrap.context.AsyncContext nextAsyncContext = recorder.recordNextAsyncContext(true);
-                if (args != null && args.length > 1 && args[0] instanceof AsyncContextAccessor) {
-                    ((AsyncContextAccessor) args[0])._$PINPOINT$_setAsyncContext(nextAsyncContext);
+                Object asyncContextAccessor = ArrayUtils.get(args, 0);
+                if (asyncContextAccessor instanceof AsyncContextAccessor) {
+                    ((AsyncContextAccessor) asyncContextAccessor)._$PINPOINT$_setAsyncContext(nextAsyncContext);
```
koo-taejin: When args has a size of 1, the old code is not executed, but changed code is going to be executed.
I don't know if the size of args can be 1.

plugins/reactor-netty/src/main/java/com/navercorp/pinpoint/plugin/reactor/netty/interceptor/HttpServerHandleHttpServerStateInterceptor.java
emeroad:
```java
      if (!(args[1] instanceof HttpResponse)) {
            // Skip LastHttpContent or else.
            return false;
        }
```
If you look at the meaning of the whole code, `getLength(args) != 1` seems to be more accurate.

jaehong-kim (approved):

Issue #7868, opened 2021-05-14
Enable/Disable pinpoint agent in java using env varibale


theLazyCat775:
Is there any env variable to enable/disable pinpoint agent in java?

For example in node, you can disable the agent by setting "PINPOINT_ENABLE=False".

itnikita:
I set up pinpoint agent like this:
 JAVA_OPTS= ”-javaagent:/usr/local/pinpoint-agent/pinpoint-bootstrap.jar -Dpinpoint.agentId=MyIP -Dpinpoint.applicationName=MyAPPName"
i think you can disable this line to disable pinpoint agent.hopes help.

BTW,do you mean in docker?

emeroad:
Resolved in https://github.com/pinpoint-apm/pinpoint/pull/7880

theLazyCat775:
Thank you @emeroad , @yjqg6666 .
This is exactly what i was looking for.

@itnikita in k8s cluster.

PR #7880, opened 2021-05-20
[#7868] disable environment/property support

yjqg6666:
# feature
To disable the pinpoint agent, even the agent is set via javaagent.

# when to use
When the javaagent jvm parameter is set as global to all applications/pods, especially for k8s environments, or set by JAVA_OPTS. Some jvm applications may need to disable the agent for now/ever/emergency.
# how to use
### 1. by environment variable
 env PINPOINT_DISABLE=true java -jar /path/to/jar
### 2. by property
java -Dpinpoint.disable=true -jar /path/to/jar


emeroad:
@yjqg6666 
Thank you for contribution

PR #7998, opened 2021-06-30
[#6617] provide percent rate sampler

### LINKED ISSUE #6617: profiler.sampling.rate restricts rate options
agent

The current supported sampling rate can only be 100%(n=1), 50%(n=2), 33%(n=3), 25%(n=4) and some other rates which are less and less.  If i wanna the sampling rate to be 80%, it can not be supported. 

The sampling rate can be n out of 100.  It is 50% when n is 50.
https://github.com/naver/pinpoint/blob/2cc86076ff58f2a2bb9273fb99c4fd3410e1e60e/profiler/src/main/java/com/navercorp/pinpoint/profiler/sampler/SamplingRateSampler.java#L42-L46   
the _isSampling_ method may need a new algorithm.

yjqg6666:
Resolve #6617 and improve PR #7281

# Changes
Provide percent rate sampler. This sampler will sample the every first x transactions out of y transactions (check the following table). The configuration may be more human-friendly.
The old/default sampler will sample the every first 1 transaction out of x transactions.

# How to enable
Disabled by default. To enable it, configure profiler.sampling.type=percentRate and update profiler.sampling.rate config accordingly. The illegal profiler.sampling.rate value is any integer value between 1 and 100.

For the performance/sampling reason, the sampling rate value which is the multiples of 2 or 5 is recommended, eg. 10, 40, 60, 75

image


yjqg6666:
This sampler can also be adapted to Per-mille rate sampler easily.

emeroad:
@yjqg6666 
Do you want a simple percent rate sampler?

How about this implementation?
```java
public class PercentRateSampler implements Sampler {

    private final Random random = new Random();
    // 1% -> 100, 100% -> 10000
    private final long percentRate = 100; 

    @Override
    public boolean isSampling() {
        int seed = random.nextInt(100 * 100);
        return percentRate > seed;
    }

}
```

emeroad:
*100 samples per second*
I think the above fixed sampling(?) is useful,
 - Algorithm for fixed sampling should be uniform distribution.

Statistics by the server should be changed to be processed by the agent
- eg) Histogram. Load, ResponeTime-avg

yjqg6666:
The suggested implementation would not ensure to have the configured sampling rate.

emeroad:
@yjqg6666  
Okay. I understand this algorithm.

I've improved it so it doesn't require a specific multiple.

```java
public class PercentRateSampler2 implements Sampler {

    private final AtomicLong counter = new AtomicLong(0);
    private final long samplingRate;
    private final long max = 100;

    public PercentRateSampler2(int samplingRate) {
        if (samplingRate <= 0 || samplingRate >= 100) {
            throw new IllegalArgumentException("Invalid samplingRate " + samplingRate);
        }
        this.samplingRate = samplingRate;
    }

    @Override
    public boolean isSampling() {
        long samplingCounter = counter.addAndGet(samplingRate);
        long remainder = MathUtils.floorMod(samplingCounter, max);
        if (remainder > 0 && remainder <= samplingRate) {
            return true;
        }
        return false;
    }

}
```

It handles all variables normally as in the test below.
```java
    @Test
    public void test2_33p() {
        PercentRateSampler2 simpleSampler5 = new PercentRateSampler2(33);
        for (int i = 0; i < 33; i++) {
            assertChoice(simpleSampler5);
            assertDrop(simpleSampler5);
            assertDrop(simpleSampler5);
        }
        assertDrop(simpleSampler5);
        assertChoice(simpleSampler5);
        assertDrop(simpleSampler5);
    }

```
33% 
| sum        |  0    |   33   |  66   |  99   | 132 |  165 |
| ---          | ---  | ---  | ---    | ---  | --- |   ---- |
| mod  100    |  0    |  33   |  66  |   99   |  32  |   65  |
| sampling | true |    X   |  X     |   X     |   true |  X    |

profiler/src/main/java/com/navercorp/pinpoint/profiler/sampler/PercentRateSampler.java
```diff
@@ -0,0 +1,104 @@
...
+    static class PercentSamplerHelper {
+
+        private int notSampledMinSeq;
+
+        private int outOfNum;
+
+        public PercentSamplerHelper(int samplingRate) {
+            this.notSampledMinSeq = samplingRate;
+            this.outOfNum = 100;
+            optimize(samplingRate);
```
yjqg6666: The "optimize" can make the sampling more evenly distributed. If the rate is set to 50%, the every first one of two requests will be sampled, instead of every first 50 requests of 100 requests. I think this optimization part is needed.

yjqg6666: @emeroad

profiler/src/main/java/com/navercorp/pinpoint/profiler/sampler/PercentRateSampler.java
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        private void optimize(int samplingRate) {
+            int numerator = samplingRate;
+            int denominator = 100;
+            boolean changed;
```
yjqg6666: As 100=10 * 10=(2 * 5)  *  (2 * 5).

emeroad:
The distribution of this algorithm is the same as before.

25%
sum | 25| 50 | 75 | 100 | 125
-- | -- | -- | -- | -- | --
mod 100 | 25 | 50 | 75 | 100 | 25
sampling | true | X | X | X | true

33%
sum | 33 | 66 | 99 | 132 | 165
-- | -- | -- | -- | -- | --
mod 100 | 33 | 66 | 99 | 32 | 65
sampling | true | X | X | true | X

50%

sum | 50 | 100 | 150 | 200 | 250
-- | -- | -- | -- | -- | --
mod 100 | 50 | 0 | 50 | 0 | 50
sampling | true | X | true  | x | true

yjqg6666:
image

emeroad:
@yjqg6666 
The modified algorithm also has the same distribution.

This algorithm remove optimize() function and increments the counter by the sampling-rate instead.
```java
@Override
    public boolean isSampling() {
        long samplingCounter = counter.addAndGet(samplingRate);  <---- Increase by samplingRate (eg: 33p, 50p)
        long remainder = MathUtils.floorMod(samplingCounter, max);
        if (remainder > 0 && remainder <= samplingRate) {
            return true;
        }
        return false;
    }
```
33%
```java
    @Test
    public void test2_33p() {
        PercentRateSampler2 simpleSampler5 = new PercentRateSampler2(33);
        for (int i = 0; i < 33; i++) {
            assertChoice(simpleSampler5);
            assertDrop(simpleSampler5);
            assertDrop(simpleSampler5);
        }  // 99th  choice:66, drop: 33
        assertDrop(simpleSampler5); <-- 100th :  choice:66, drop:33 + 1 = 100
        assertChoice(simpleSampler5);
        assertDrop(simpleSampler5);
    }
```

yjqg6666:
@emeroad Could it pass the updated tests in this PR?

yjqg6666:
@emeroad I get you now and samplingCounter might need fastAbs operation due to overflow.

emeroad:
image
@yjqg6666  
Yes, the modified algorithm passes the above tests.
Even for complex cases like 33% (choosing 30.030303... out of 100), this algorithm has a good distribution.

PR #8135, opened 2021-08-19
[#8133] show Avg&Max Response summary info on ServerMap edge

### LINKED ISSUE #8133: show Avg&Max Response summary info on ServerMap edge
UI

A PR would be made.

yjqg6666:
Resolve #8133.

# Feature
In main server map:
image

In transaction server map:
image

### DIFF
--- a/web/src/main/angular/src/app/core/components/server-map/class/server-map-diagram-with-cytoscapejs.class.ts
+++ b/web/src/main/angular/src/app/core/components/server-map/class/server-map-diagram-with-cytoscapejs.class.ts
@@ -275,6 +275,8 @@ export class ServerMapDiagramWithCytoscapejs extends ServerMapDiagram {
                 map(([nodes, _]: {[key: string]: any}[][]) => {
                     const edges = addedEdgeList.map((edge: {[key: string]: any}) => {
                         const {from, to, key, totalCount, hasAlert, isMerged} = edge;
+                        const originalLinkData = serverMapData.getLinkData(key);
+                        const responseInfo = this.getResponseInfo(originalLinkData);
 
                         return {
                             data: {
@@ -282,7 +284,7 @@ export class ServerMapDiagramWithCytoscapejs extends ServerMapDiagram {
                                 source: from,
                                 target: to,
                                 isMerged,
-                                label: totalCount.toLocaleString(),
+                                label: `${totalCount.toLocaleString()}${responseInfo}`,
                                 hasAlert,
                                 alive: true
                             }
@@ -313,6 +315,8 @@ export class ServerMapDiagramWithCytoscapejs extends ServerMapDiagram {
             const edgeList = serverMapData.getLinkList();
             const edges = edgeList.map((link: ILinkInfo) => {
                 const {from, to, key, totalCount, isFiltered, isMerged, hasAlert} = link;
+                const originalLinkData = serverMapData.getLinkData(key);
+                const responseInfo = this.getResponseInfo(originalLinkData);
 
                 return {
                     data: {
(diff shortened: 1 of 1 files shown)


binDongKim:
We have talked through this feature and it seems like a decent one but at the same time, if the server-map is complicated, things would become harder to recognize with this look. So I guess we will need a better but simpler way to display the info such as popping-up a small info-box when hovering the mouse on the link. 

Thank you.

yjqg6666:
On the right sidebar, there is already avg&max chart which is shown only when clicking on edges. Popping-up is still indirect and unnecessary. How about showing different line thickness and/or color regarding avg&max value range?

binDongKim:
That could work but just wondering how it would look like with that approach.
Would you try implementing it with a middle-big size server-map and share it here if it doesn't look confusing?

emeroad:
@yjqg6666 
Can you remove MAX and put only AVG?
It would be nice to see something like UserLink `1 (9006)`

Data Format
1. 9999 (100)
2. 9999 / 100

Would 1 be better?

emeroad:
In my opinion, it would be better to add TimeUnit to AVG

100  -> 100 ms
6000 -> 1.0 sec
6100 -> 1.1 sec
61000 -> 6.1 min

There may be a way to omit precision.
6002 -> 1.0 sec

yjqg6666:
Update:

In the main server map:
image
In the transaction server map:
image

web/src/main/angular/src/app/core/components/server-map/class/server-map-diagram-with-cytoscapejs.class.ts
```diff
@@ -275,14 +275,16 @@ export class ServerMapDiagramWithCytoscapejs extends ServerMapDiagram {
...
+                        const responseInfo = this.getResponseInfo(originalLinkData);
 
                         return {
                             data: {
                                 id: key,
                                 source: from,
                                 target: to,
                                 isMerged,
-                                label: totalCount.toLocaleString(),
+                                label: totalCount.toLocaleString() + responseInfo,
```
binDongKim: Use template literals when concatenating strings.
```suggestion
                                label: `${totalCount.toLocaleString()}${responseInfo}`,
```

web/src/main/angular/src/app/core/components/server-map/class/server-map-diagram-with-cytoscapejs.class.ts
```diff
@@ -323,7 +327,7 @@ export class ServerMapDiagramWithCytoscapejs extends ServerMapDiagram {
                         // [임시]label에서 이미지를 지원하지않아서, filteredMap페이지에서 필터아이콘을 "Filtered" 텍스트로 대체.
                         // label: isFiltered ? ` [Filtered]\n${totalCount.toLocaleString()} ` : ` ${totalCount.toLocaleString()} `,
                         // TODO: Filter Icon 처리
-                        label: totalCount.toLocaleString(),
```
binDongKim, line 326: Same as above

web/src/main/angular/src/app/core/components/server-map/class/server-map-diagram-with-cytoscapejs.class.ts
```diff
@@ -346,6 +350,27 @@ export class ServerMapDiagramWithCytoscapejs extends ServerMapDiagram {
...
+    private formatTime(val: number): string {
+        if (val < 1000) {
+            return val + ' ms';
+        }
+        if (val % 60000 == 0) {
+            return (val / 60000) + ' min';
+        }
+        return (val / 1000.0).toFixed(2).replace('.00', '').replace(/(\.\d)0$/, '$1') + ' sec';
+    }
+
```
binDongKim: 1. Add line-break after `if` statement and variable declaration.
2. Use template literals when concatenating strings.
3. Use conditional ternary operator to make the code concise.

```suggestion
    getResponseInfo(linkData: ILinkInfo|any): any {
        if (typeof linkData === 'undefined' || !linkData) {
            return '';
        }

        const histogramAvg = linkData.responseStatistics ? linkData.responseStatistics.Avg : 0;

        return histogramAvg > 0 ? ` (${this.formatTime(histogramAvg)})` : '';
    }

    private formatTime(val: number): string {
        return val < 1000 ? `${val} ms`
            : val % 60000 === 0 ? `${(val / 60000)} min`
            : `${(val / 1000.0).toFixed(2).replace('.00', '').replace(/(\.\d)0$/, '$1')} sec`;
    }
```

binDongKim (changes requested):
I've left a few comments mostly about the syntax. 

Check them out and after you change them, please rebase(squash) all the commits into one.

Thank you.

yjqg6666:
@binDongKim Squashed and rebased.

PR #8163, opened 2021-08-27
[#8162] cache applications api on browser/server side

### LINKED ISSUE #8162: [FR] Cache applications.pinpoint api on browser/server side for quicker page loading.
Web

eTag + ehcache

PR #8163  has been made.

yjqg6666:
Resolve #8162 

# Feature
By default, the application list would be cached by Ehcache for 120 seconds(configured in ehcache.xml) and be cached by browser  for 60 seconds (configured by web.applicationList.cacheTime in pinpoint-web-root.properties). If the request having no "If-Non-Match" header(when it's the first time to request or after expired browser cache time), ehcache would be evicted and retrieve data from hbase. The cache can also be clear by requesting {web_host}/applications.pinpoint?clearCache manually.
**The first time to request for a user, there would be no cache in use.**

You may let the web browser cache for a longer time by setting `-Dweb.applicationList.cacheTime` to a bigger value (in seconds). The following request would get "304 Not Modified response header" and no response body. The browser would use previously cached data.

image

In the above picture:
1. no cache on the browser & server sides.
2. cache on the browser side.
3. cache on the server side only and no application list data to transfer.

# Verification
```
# request for the first time or with expired cache
$ curl 'http://localhost:8080/applications.pinpoint' -D-   

# request with eTag returned by the previous response eTag header value
$ curl 'http://localhost:8080/applications.pinpoint' -D- -H 'If-None-Match: "applicationNameListVer1"'  

# clear server cached list
$ curl 'http://localhost:8080/applications.pinpoint?clearCache' -D-

```

### DIFF
--- a/web/src/main/angular/src/app/core/components/application-list/application-list-data.service.ts
+++ b/web/src/main/angular/src/app/core/components/application-list/application-list-data.service.ts
@@ -15,8 +15,10 @@ export class ApplicationListDataService {
         private http: HttpClient,
     ) {}
 
-    getApplicationList(): Observable<IApplication[]> {
-        return this.http.get<IApplication[]>(this.url).pipe(
+    getApplicationList(force: boolean): Observable<IApplication[]> {
+        const options = force ? {params: {clearCache:'true', from: 'reload', _t: Date.now().toString()}} : {};
+
+        return this.http.get<IApplication[]>(this.url, options).pipe(
             // TODO: 워커적용?
             map((res: IApplication[]) => res.map(({applicationName, serviceType, code}) => new Application(applicationName, serviceType, code))),
         );
--- a/web/src/main/java/com/navercorp/pinpoint/web/controller/MainController.java
+++ b/web/src/main/java/com/navercorp/pinpoint/web/controller/MainController.java
@@ -16,23 +16,25 @@
 
 package com.navercorp.pinpoint.web.controller;
 
-import com.navercorp.pinpoint.common.util.StringUtils;
 import com.navercorp.pinpoint.web.service.CacheService;
 import com.navercorp.pinpoint.web.service.CommonService;
+import com.navercorp.pinpoint.web.util.etag.ETag;
+import com.navercorp.pinpoint.web.util.etag.ETagUtils;
 import com.navercorp.pinpoint.web.view.ApplicationGroup;
 import com.navercorp.pinpoint.web.view.ServerTime;
+import com.navercorp.pinpoint.web.view.TagApplications;
 import com.navercorp.pinpoint.web.vo.Application;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RequestHeader;
+import org.springframework.web.bind.annotation.RequestParam;
 import org.springframework.web.bind.annotation.RestController;
 
 import java.util.List;
 import java.util.Objects;
-import org.springframework.http.ResponseEntity;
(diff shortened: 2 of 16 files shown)


web/src/main/java/com/navercorp/pinpoint/web/controller/MainController.java
```diff
@@ -39,21 +44,41 @@
...
-        return new ApplicationGroup(applicationList);
+        final String newETag = cacheService.getApplicationListETag();
+        final boolean noUpdate = eTag != null && eTag.replace("\"", "").contentEquals(newETag);
+        final ResponseEntity.BodyBuilder builder = ResponseEntity.ok().eTag(newETag);
+        if (noUpdate) {
+            return builder.contentLength(0L).build();
+        } else {
+            return builder.body(new ApplicationGroup(applicationList));
+        }
     }
```
emeroad, line 133: @yjqg6666 
In case of `noUpdate`, it seems that NOT_MODIFIED(304) should be set in HttpStatus.
```java
        if (noUpdate) {
            final ResponseEntity.BodyBuilder builder = ResponseEntity.status(HttpStatus.NOT_MODIFIED).eTag(newETag);
            return builder.build();
        } else {
            ResponseEntity.BodyBuilder builder = ResponseEntity.ok().eTag(newETag);
            return builder.body(new ApplicationGroup(applicationList));
        }
```

yjqg6666: I tested and got status code 304, so no status code set explicitly then. I will add it soon.

emeroad:
How about changing the applicationList's reload button click action?
- Bypass browser cache and force update applicationlist for data freshness

eg)
1. Add magic key to /applications.list
2. Add clearCache parameter to applications.list

/applications.list?clearCache&magic=1234

yjqg6666:
@emeroad  great idea.

yjqg6666:
@emeroad implemented.

image

binDongKim:
@yjqg6666 
When I click the reload button, it passes the parameters which I guess is for bypassing the cache and update the data but I noticed the browser keeps fetching the data from the cache even when clicking the button in short period.

스크린샷 2021-09-03 오후 2 43 38

Is this an expected behavior?

yjqg6666:
@binDongKim It's not and a new commit is appended to fix it.

web/src/main/angular/src/app/core/components/application-list/application-list-data.service.ts
```diff
@@ -15,8 +15,9 @@ export class ApplicationListDataService {
         private http: HttpClient,
     ) {}
 
-    getApplicationList(): Observable<IApplication[]> {
-        return this.http.get<IApplication[]>(this.url).pipe(
+    getApplicationList(force: boolean): Observable<IApplication[]> {
+        const options = force ? {params: {clearCache:'true', from: 'reload', _t: Date.now().toString()}} : {};
+        return this.http.get<IApplication[]>(this.url, options).pipe(
```
binDongKim:
```suggestion
        const options = force ? {params: {clearCache:'true', from: 'reload', _t: Date.now().toString()}} : {};
        
        return this.http.get<IApplication[]>(this.url, options).pipe(
```
Please add a line-break after declaration.

binDongKim (changes requested):
Other than the simple line-break comment, everything seems to work well.

After you change it, please rebase and squash `c00443d75186be4487d62a770c14e8753cfd2bc6`,
`9152148ad83f03929a39045d984aca207cf7a091` commits into one commit.

yjqg6666:
@binDongKim Done and thanks for the guide and review.

binDongKim:
@yjqg6666 
Things look good to go in frong-end side. Thank you!

binDongKim (approved):

emeroad (approved):

PR #8167, opened 2021-08-30
[#8166] disable sampling flag header can be ignored

### LINKED ISSUE #8166: [FR.] disable sampling flag header can be ignored (configurable)
Profiling

What are the considerations when the sampling flag was introduced? From the issues and commit history i haven't found any info about it.

If the topo is: gatewayApp (50% sampling) -> bizApp1 (100% profiling) -> bizApp2, the sampling rate for the gatewayApp is 50% and bizApp1 is 100%. The unsampled request through gatewayApp will add pinpoint header "Pinpoint-Sampled: s0". When the bizApp1 receive the request, it will recognize the header and decide not to trace the request, causing the bizApp1 not 100% profiling. 

** Solution
We might make the trace header reader ignore the header and treat it as a new trace and make this behavior configurable, default to not ignore the header for BC.  It's open to discussion.

yjqg6666:
Resolve #8166.

This feature is off by default, can be enabled  by setting profiler.sampling.ignoreDisableSamplingFlag=true.


yjqg6666:
rebased on the current master.

yjqg6666:
Rebased on the current master.

PR #8170, opened 2021-08-30
[#8169] still send caller application info when not profiled

### LINKED ISSUE #8169: [FR.] send caller application info regardless of sampling rate
agent

When a request is not sampled, we can still capture(tcpdump)/record the request header Pinpoint-pAppName to know who is calling my app. If both applications have different sampling rate, the caller may be identified by "User", if we send the caller application info and use the recording request header feature, then we could know who is calling from a call tree.

A PR #8170 is made to resolve this issue.

yjqg6666:
Resolve #8169.

This feature is disabled by default for BC. It can be enabled by agent config 'profiler.sendAppInfo.enable=true'.

These headers are sent:
- Pinpoint-pAppName
- Pinpoint-pAppType
- Pinpoint-pAppNamespace (only when it's set)

# tasks

- [x] http client support
- [x] kafka client support (only if kafka client/broker support header & profiler.kafka.header is enabled and profiler.kafka.sendAppInfo.enable=true)

### DIFF
--- a/agent/src/main/resources/pinpoint-root.config
+++ b/agent/src/main/resources/pinpoint-root.config
@@ -332,6 +332,11 @@ profiler.http.status.code.errors=5xx
 # e.g. profiler.http.record.response.headers=Set-Cookie
 #profiler.http.record.response.headers=
 
+# send caller application info even when the request is not profiled
+#profiler.sendAppInfo.enable=false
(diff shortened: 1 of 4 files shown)


yjqg6666:
Closed. PR #8561 is preferred.

PR #8352, opened 2021-10-27
[#8341] It is possible to specify the znode path of zookeeper used in…

koo-taejin:
… cluster

1. Apply ZNode path config to ZookeeperClusterConfiguration
2. Support ZNode path config for collector and web module


collector/src/main/resources/pinpoint-collector-root.properties
```diff
@@ -141,5 +142,6 @@ collector.map-link.max.enable=true
 # Flink configuration
 flink.cluster.enable=false
 flink.cluster.zookeeper.address=${pinpoint.zookeeper.address}
+flink.cluster.zookeeper.znode_root=/pinpoint-cluster
 flink.cluster.zookeeper.path=/pinpoint-cluster/flink
```
minwoo-jung: You can delete this line.:)

koo-taejin: I have resolved, thanks for sharing this.

PR #8375, opened 2021-11-05
[#8374] show date of start time as a tooltip in web call tree

### LINKED ISSUE #8374: [FR.] show date of starttime in call tree
UI

From a transaction info page, there is no way to know the start day of the transaction. The date is shown as a tooltip.
After:
image

Before:
image

A PR #8375 is made to resolve this issue.

Alternative: the date info can also be shown as a part of the cell/field, dateformat need update.

yjqg6666:
Resolve #8374.

## Feature:

Before:
image

After:
image


binDongKim:
We were discussing that if the date info is really necessary because it's already provided in the upper part of the page.
스크린샷 2022-01-05 오후 12 21 01

yjqg6666:
The use case is that we get the transaction id from application log/response header, and search/view the transaction info by TxId. We only get some pages like this: http://125.209.240.10:10123/transactionDetail?transactionInfo=%7B%22agentId%22:%22apigw01%22,%22spanId%22:%22-1%22,%22traceId%22:%22apigw01%5E1640327922199%5E4092935%22,%22collectorAcceptTime%22:%220%22%7D . There is no date info.
image

PR #8624, opened 2022-01-27
Add support 5.6.0 ~ 7.4.x RestClient to elasticsearch plugin

aalinyu:
1. Add support 5.6.0 ~ 7.4.x RestClient to elasticsearch plugin. HighLevelClient is base on RestClient,so it is also support HighLevelClient 5.6.0 ~ 7.4.x
2. Add elastic index and method Annotation, The elastic api meet restful standards,so we can get index from rest url.
3. The plugin was tested and run in production environment 
image


aalinyu:
@emeroad Any thought/suggestion on this feature?

PR #8860, opened 2022-05-12
[#noissue] add isExistUserGroupMember

smilu97:
Added `boolean isExistUserGroupMember(String userGroupId, String memberId)` in UserGroupDao.


minwoo-jung (approved):
good!

PR #8874, opened 2022-05-20
Add HTable method to HBase Plugin

jimolonely:
Distinguishing HBase read and write requests are helpful to analyze the usage of APP behavior.
So I record the HTable method to realize the function.

PR #8875, opened 2022-05-20
Add HBase HTable method dataSize intercept to HBase plugin

jimolonely:
One step further from [#8874 ](https://github.com/pinpoint-apm/pinpoint/pull/8874), we want to record the client request data size, including the write method and read method.
we calculate the parameter data size of the write method, like `put(Get)`, and the result of the read method, like `Result get(Get)`.
We desire to calculate the scan result data size, but it affects performance severely, we have to compromise, only recording the RPC result number. 
Hope your suggestions!


intr3p1d:
### Info
This feature enables user to trace the data size (approximated by the argument result of Hbase client methods)
The image below is an example.


image

Issue #8883, opened 2022-05-25
Fix return incorrect values when executing getAsyncContext() in Reactor plugin


koo-taejin:
#### description 
```java
    public AsyncContext getAsyncContext(Object target, Object[] args) {
        if (checkTargetReactorContextAccessor(target, args)) {
            return AsyncContextAccessorUtils.getAsyncContext(target); // -> return ReactorContextAccessorUtils.getAsyncContext(target);
        }
        if (checkTargetAsyncContextAccessor(target, args)) {
            return AsyncContextAccessorUtils.getAsyncContext(target);
        }
        if (checkSubscriberReactorContextAccessor(target, args)) {
            return AsyncContextAccessorUtils.getAsyncContext(target); // return ReactorContextAccessorUtils.getAsyncContext(target);
        }
        return AsyncContextAccessorUtils.getAsyncContext(target);
    }

```

PR #8884, opened 2022-05-25
[#8883] Fix return incorrect values when executing getAsyncContext() …

koo-taejin:
…in Reactor plugin


koo-taejin:
@jaehong-kim 

I think this part also seems to need to be checked.
https://github.com/pinpoint-apm/pinpoint/blob/v2.4.0-RC1/bootstraps/bootstrap-core/src/main/java/com/navercorp/pinpoint/bootstrap/plugin/reactor/FluxAndMonoSubscribeInterceptor.java#L94

In the after() method, the AsyncContext is being fetched via AsyncContextAccessorUtils.
If AsyncContext is saved via with ReactorContextAccessorUtils in before(), it cannot be found after, it could be occurred corrupted.

It seems necessary to unify the way to search AsyncContext or there seems to be a way to fix the storage location.

Thanks :)

jaehong-kim (approved):

jaehong-kim:
@koo-taejin 
Thank you.

PR #8961, opened 2022-06-22
[#noissue] Support tls certification chain

smilu97:
Currently, `GrpcSslConfiguration`, and `SslServerConfig` can't contain multiple certification file, but only support utilizing a self-signed certification.

These configurations are mostly for arguments of `SslContextBuilder.forServer` ([javadoc](https://netty.io/4.0/api/io/netty/handler/ssl/SslContextBuilder.html)), but there is no way to call `forServer(java.security.PrivateKey key, java.lang.String keyPassword, java.security.cert.X509Certificate... keyCertChain)`

### DIFF
--- a/collector/src/main/java/com/navercorp/pinpoint/collector/grpc/config/GrpcSslConfiguration.java
+++ b/collector/src/main/java/com/navercorp/pinpoint/collector/grpc/config/GrpcSslConfiguration.java
@@ -30,14 +30,16 @@ public class GrpcSslConfiguration {
     private final boolean enable;
     private final String providerType;
     private final Resource keyResource;
-    private final Resource keyCertChainResource;
+    private final Resource[] keyCertChainResources;
+    private final String keyPassword;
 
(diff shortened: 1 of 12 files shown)


collector/src/main/java/com/navercorp/pinpoint/collector/grpc/config/GrpcSslConfiguration.java
```diff
@@ -128,7 +148,18 @@ public String toString() {
                 "enable=" + enable +
                 ", providerType='" + providerType + '\'' +
                 ", keyResource='" + keyResource + '\'' +
-                ", keyCertChainResource='" + keyCertChainResource + '\'' +
+                ", keyCertChainResource='" + joinResources(keyCertChainResources) + '\'' +
```
emeroad, line 151: Arrays.toString() recommended.

emeroad (changes requested):

smilu97:
Current method is already able to cover common usages of TLS certifications.

PR #9035, opened 2022-07-11
[#noissue] modified incorrect word in document alarm part 임게치 to 임계치

devdynam0507:
modified incorrect word in document alarm part word `임게치` to `임계치`


minwoo-jung:
@devdynam0507 
Thank you for your contribution. ^^

But
We do not make modifications to previous versions of the document. 
It is a very good idea to maintain the past version.
But unless there is a special case, patch versions are not released.
so even if there are typos in the past documents,
they are usually not reflected and only the documents of the master branch are maintained.

minwoo-jung:
@devdynam0507 
If you would like to contribute to the documentation,
we recommend that you check out the documentation in the https://github.com/pinpoint-apm/pinpoint-apm.github.io repository.

Issue #9078, opened 2022-07-27
Deletes should be committed frequently when cleaning inactive agents


smilu97:
Finding all inactive agents may consume enormous time, but current `CleanupInactiveAgentJob` is too sensitive to any unpredictable error which is caused by outer components (ex. hbase, tcp)

PR #9088, opened 2022-08-03
[#9078] Stable cleanup

smilu97:
`CleanupInactiveAgentsJob` is enormous, and takes about 10h per execution.

* It collects dead agents by traversing all agents, and frequently delete agents when 100 dead agents are found.
* `CleanupInactiveAgentsTasklet` split chunks more granuly for preventing transaction-timeout-error

### DIFF
--- a/web/src/main/java/com/navercorp/pinpoint/web/service/AdminServiceImpl.java
+++ b/web/src/main/java/com/navercorp/pinpoint/web/service/AdminServiceImpl.java
@@ -16,25 +16,17 @@
 
 package com.navercorp.pinpoint.web.service;
 
+import com.navercorp.pinpoint.common.server.util.time.Range;
 import com.navercorp.pinpoint.common.util.CollectionUtils;
 import com.navercorp.pinpoint.web.dao.ApplicationIndexDao;
 import com.navercorp.pinpoint.web.vo.Application;
-import com.navercorp.pinpoint.common.server.util.time.Range;
-
-import org.apache.logging.log4j.Logger;
 import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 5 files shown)


web/src/main/java/com/navercorp/pinpoint/web/service/AdminServiceImpl.java
```diff
@@ -147,30 +177,44 @@ private List<String> filterInactiveAgents(List<String> agentIds, int durationDay
...
-            // If at any time a non-java agent is introduced, or an agent that does not collect jvm data,
-            // this will fail
+        return !this.agentInfoService.isActiveAgent(agentId, range);
+    }
 
-            boolean dataExists = agentInfoService.isActiveAgent(agentId, fastRange);
-            if (dataExists) {
-                continue;
+    @NotThreadSafe
+    private static class BulkConsumer<T> implements Consumer<T> {
```
minwoo-jung: It seems that the code is difficult to understand because consumer has another consumer.

web/src/main/java/com/navercorp/pinpoint/web/service/AdminServiceImpl.java
```diff
@@ -71,26 +65,62 @@ public void removeInactiveAgents(int durationDays) {
...
+    }
+
+    private void waitOneMinute() {
+        try {
+            Thread.sleep(60000);
+        } catch (Exception ignored) {}
+    }
+
+    private int removeInactiveAgentInApplication0(String applicationName, int durationDays) {
+        final BulkConsumer<String> remover = new BulkConsumer<>(agentIds -> {
```
minwoo-jung: It seems that the code is difficult to understand because consumer has another consumer.

minwoo-jung (commented):
good!

PR #9158, opened 2022-08-26
[#9157] Prevent initializing active thread workers multiple times

### LINKED ISSUE #9157: Only one user can use realtime-active-request feature on same machine and application

smilu97:
resolves #9157


emeroad (approved):
LGTM

PR #9275, opened 2022-10-18
fix(sec): upgrade com.ning:async-http-client to 1.9.0

claire9910:
### What happened？
There are 1 security vulnerabilities found in com.ning:async-http-client 1.8.3
- [CVE-2013-7398](https://www.oscs1024.com/hd/CVE-2013-7398)

### What did I do？
Upgrade com.ning:async-http-client from 1.8.3 to 1.9.0 for vulnerability fix

### What did you expect to happen？
Ideally, no insecure libs should be used.

### The specification of the pull request
[PR Specification](https://www.oscs1024.com/docs/pr-specification/) from OSCS


emeroad:
Plugin dependencies only affect tests. There is no security impact.

Issue #9315, opened 2022-10-24
add metric chart order


donghun-cho:
sort collectedMetricInfo return value in yml order

PR #9316, opened 2022-10-24
[#9315] Add order in collected metric info, metric host info

donghun-cho:
sort metric host info
sort collected metric in telegraf-metric.yml order


minwoo-jung (approved):
good!

PR #9372, opened 2022-11-03
[#9371] Add agent load-balancer which periodically change collector server

### LINKED ISSUE #9371: Add agent load-balancer which periodically change collector server
SendSpan takes lots of portion of collector usage (maybe).

Long-live streams are generally requires to be renewed periodically for maintaining good balance of load in backend of L4-proxy.

This client-side load balancer should be able to update the available transports properly. For now experimentally, it implements a very simple load balancer which just periodically disconnects the old transport, and creates a new one.

smilu97:
resolves #9371

### DIFF
--- a/agent/src/main/resources/pinpoint-root.config
+++ b/agent/src/main/resources/pinpoint-root.config
@@ -19,6 +19,9 @@ profiler.transport.module=GRPC
 ###########################################################
 profiler.transport.grpc.collector.ip=127.0.0.1
 
+# subconnection expiring loadbalancer
+profiler.transport.grpc.loadbalancer.renew.period.millis=3153600000000
+
 # placeHolder support "${key}"
 # Agent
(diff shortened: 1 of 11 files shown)


profiler/src/main/java/com/navercorp/pinpoint/profiler/grpc/SubconnectionExpiringLoadBalancer.java
```diff
@@ -0,0 +1,247 @@
...
+            final AtomicReference<PickProgress> progress = getPickProgress(subchannel);
+            if (progress.compareAndSet(PickProgress.NOT_PICKED_YET, PickProgress.PICKED)) {
+                helper.getSynchronizationContext().execute(this::createSubchannel);
+            }
+        }
+    }
+
+    @Override
+    public void requestConnection() {
+        this.readySubchannel.requestConnection();
```
smilu97: TODO: call `createSubChannel` instead

PR #9383, opened 2022-11-07
[#noissue] Replace hamcrest into assertj

smilu97:


smilu97:
Changed branch-name, #9389 #9390

Issue #9414, opened 2022-11-14
PinpointPluginTestSuite JDK9+ compatibility


smilu97:
As referred at #7415, ITs are currently only compatible on JDK8.

PR #9416, opened 2022-11-15
[#9414] Refer classpath for querying available libraries

smilu97:


smilu97:
All changes hereare adapted from #7415, but what was in `ModuleSupport` is excluded because the issue had been already resolved by the other PR

Issue #9424, opened 2022-11-20
Export the 'mail.smtp.ssl.enable' in the batch.


JunRadish:
**Which module is your feature request related to?**
Batch/.../applicationContext-batch-sender.xml

**Is your feature request related to a problem?**
Cannot use Gmail's SMTP easily cause by default, the sender does not enabled to use 'mail.smtp.ssl.enable' configuration.
Cause a lot of users in pinpoint might using gmail to send mail via pinpoint's batch, this feature will be useful to lots of develpers!

**Describe the solution you'd like**
1. Add the below property between line number 30 and 31 in 'pinpoint/batch/src/main/resources/applicationContext-batch-sender.xml'
`<prop key="mail.smtp.ssl.enable>${alarm.mail.smtp.ssl.enable:false}</prop>
`
2. Update the configuration guide of pinpoint batch with 'Setting the value `alarm.mail.smtp.ssl.enable` if you needed'.

**Describe alternatives you've considered**
Nothing for now

**Additional context**
Thanks for sharing the most useful APM tool for me.
Me and my team always thanks to your team.
추가적으로 문의사항이 있으시다면 junminkim@jobis.co로 연락주시면 감사하겠습니다!

emeroad:
I completely agree with your proposal.
Coud you create a PR?

JunRadish:
@emeroad Sure! I will create a PR for this Issue.

PR #9441, opened 2022-11-23
[#9424] Export the 'mail.smtp.ssl.enable' in the batch. 

JunRadish:
Which module is your feature request related to?
Batch/.../applicationContext-batch-sender.xml

Is your feature request related to a problem?
Cannot use Gmail's SMTP easily cause by default, the sender does not enabled to use 'mail.smtp.ssl.enable' configuration.
Cause a lot of users in pinpoint might using gmail to send mail via pinpoint's batch, this feature will be useful to lots of develpers!

Describe the solution you'd like

Add the below property between line number 30 and 31 in 'pinpoint/batch/src/main/resources/applicationContext-batch-sender.xml'
<prop key="mail.smtp.ssl.enable>${alarm.mail.smtp.ssl.enable:false}</prop>
Update the configuration guide of pinpoint batch with 'Setting the value alarm.mail.smtp.ssl.enable if you needed'.
Describe alternatives you've considered
Nothing for now

Additional context
Thanks for sharing the most useful APM tool for me.
Me and my team always thanks to your team.
추가적으로 문의사항이 있으시다면 [junminkim@jobis.co](mailto:junminkim@jobis.co)로 연락주시면 감사하겠습니다!

### DIFF
--- a/batch/src/main/resources/applicationContext-batch-sender.xml
+++ b/batch/src/main/resources/applicationContext-batch-sender.xml
@@ -30,6 +30,7 @@
                 <prop key="mail.smtp.auth">${alarm.mail.smtp.auth:false}</prop>
                 <prop key="mail.smtp.starttls.enable">${alarm.mail.smtp.starttls.enable:false}</prop>
                 <prop key="mail.smtp.starttls.required">${alarm.mail.smtp.starttls.required:false}</prop>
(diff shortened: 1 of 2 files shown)


JunRadish:
I export the 'mail.smtp.ssl.enable' from the sender xml file and added the prop in the batch-root.properties file.
Still update is needed on the document to user's can use the prop.

emeroad:

Issue #9380, opened 2022-11-04
Linenumber 


feelform:
Line number in Source code
API Object nullability
<img width="559" alt="Screenshot 2022-11-29 at 3 28 20 PM" src="https://user-images.githubusercontent.com/616895/204455252-b31c54b6-48ce-4bf6-8849-13dc32710d2e.png">

API Object with Line number and location
<img width="841" alt="Screenshot 2022-11-29 at 3 32 18 PM" src="https://user-images.githubusercontent.com/616895/204456095-a5d76910-a26d-4680-bf6e-9e48ba742dfe.png">

Add Line number and location in APIMetaDataBO
If Line number is null, value `0`
location value is nullable
* Warning:(91, 29) 'StringBuilder sb' can be replaced with 'String'
* Hbase must save data with null because of saved data order
* Add linenumber and location in TransactionInfoViewModel

PR #9450, opened 2022-11-30
[#9380] Supports Line number and location in Callstack

feelform:
Add Line number and location in APIMetaDataBO
If Line number is null, value `0`
location value is nullable

ref #9380


emeroad (approved):
LGTM

PR #9483, opened 2022-12-07
[#9482] Await termination of grpc server

### LINKED ISSUE #9482: Enhance graceful shutdown
GrpcReceiver should wait until termination of grpc-server before closing thread pool executors.

Currently, it is possible for grpc-server to request critical jobs to executors which is already closed by destroying process of spring bean

smilu97:
resolves #9482

### DIFF
--- a/collector/src/main/java/com/navercorp/pinpoint/collector/receiver/grpc/GrpcReceiver.java
+++ b/collector/src/main/java/com/navercorp/pinpoint/collector/receiver/grpc/GrpcReceiver.java
@@ -34,8 +34,8 @@ import io.grpc.ServerCallExecutorSupplier;
 import io.grpc.ServerInterceptor;
 import io.grpc.ServerServiceDefinition;
 import io.grpc.ServerTransportFilter;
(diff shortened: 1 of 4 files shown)


collector/src/main/java/com/navercorp/pinpoint/collector/receiver/grpc/GrpcReceiver.java
```diff
@@ -159,15 +160,37 @@ private void addService() {
...
+        }
+
+        server.shutdownNow();
+        awaitServerTermination(1000);
+    }
+
+    private boolean awaitServerTermination(long maxWaitTimeMillis) {
+        try {
+            return server.awaitTermination(maxWaitTimeMillis, TimeUnit.MILLISECONDS);
+        } catch (InterruptedException ignored) {}
```
emeroad:
```
catch (InterruptedException e) {
            Thread.currentThread().interrupt();
}
```
I think it would be better to fix it to call `Thread.interrupt()`.

PR #9499, opened 2022-12-09
[#noissue] Apply GuardLogStatement for classes that execute a lot

koo-taejin:
While analyzing webflux application performance issues, I found that some pinpoint logs use a lot of CPU.
To reduce CPU usage, GuardLogStatement has been applied to classes that execute a lot.

pinpoint_log_trace

Thanks.  :)


emeroad:

Issue #9614, opened 2023-01-09
Apply temporary fix to Pinot #6834 issue


ga-ram:
https://github.com/apache/pinot/issues/6834

PR #9615, opened 2023-01-09
[#9614] Apply temporary fix to Pinot #9614 issue

ga-ram:
This fix still doesn't handle the case in which query itself has fixed parameters which includes `?`.
To handle that case `fillStatementWithParameters` also has to check if `?` is within `'`.


emeroad:
Resolved #9592

Issue #9653, opened 2023-01-25
add PluginPackageClassRequirementFilter


donghun-cho:
check required class before defining plugin class from jar

for some plugin using both javax.servlet-api and jakarta.servlet-api

PR #9658, opened 2023-01-26
[#9653] Add plugin package class requirement filter

donghun-cho:


profiler/src/main/java/com/navercorp/pinpoint/profiler/instrument/classloading/PlainClassLoaderHandler.java
```diff
@@ -107,6 +107,9 @@ private boolean isPluginPackage(String className) {
         return pluginConfig.getPluginPackageFilter().accept(className);
     }
 
+    private boolean isPluginPackageRequirementMeet(String className) {
```
intr3p1d: How about `meetsPluginPackageRequirement` instead of `is~` ?

smilu97: `satisfyingPackageRequirements` also

donghun-cho: thanks, I will use `meetsRequirement`

PR #9705, opened 2023-02-16
[#noissue] Handle ByteString annotation values of protobuf

kojandy:
Since `bytes` in protobuf model corresponds to `com.google.protobuf.ByteString` instead of `byte[]` in Java, I've added logic to convert `ByteString` to `byte[]` when `ByteString`-typed annotation value is received.

### DIFF
--- a/commons-server/src/main/java/com/navercorp/pinpoint/common/server/bo/AnnotationFactory.java
+++ b/commons-server/src/main/java/com/navercorp/pinpoint/common/server/bo/AnnotationFactory.java
@@ -59,6 +59,7 @@ public class AnnotationFactory<T> {
         } else if (value instanceof Double) {
             return value;
         } else if (value instanceof byte[]) {
+            // not supported by protobuf
             return value;
         }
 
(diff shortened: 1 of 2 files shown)


smilu97:
I'm sorry that I didn't realized it last time, but I'm not sure if it's ok to make `AnnotationFactory` grpc-dependent...?

I think we can make another `GrpcAnnotationFactory` which extends `AnnotationFactory`, and use it in `GrpcSpanBinder`

kojandy:
Yup, as @emeroad already pointed it out on last weekly meeting, I will move relating-logic to the location you've suggested. and I guess this PR won't be merged until 2.6.0 release when the actual use of the type `bytes` in protobuf model is introduced, right? Or can I merge this PR regardless of the milestone label?

smilu97:
I didn't catch that conversation was about this XD

Anyway, the milestone labels are maybe just for some summarizing stuffs like release note, and pointing out the features related to those commits are experimental. I think it is totally fine to merge some branches now, and do some experiments on the dev cluster

kojandy:
Oh, I see. Thank you for the guideline. I will fix it right away then!

smilu97:
Sorry for misguiding.

Setting milestone means the PR requires to be merged after the version exactly as you said.

PR #9938, opened 2023-05-16
[#noissue]Add recoder HTTP messages

wangjiee1:
[#noissue]Add recoder HTTP messages

The collection of HTTP messages has been implemented, including but not limited to request headers, request bodies, response headers, response bodies, request URLs, etc.

HTTP message collection data is sent through the spanWebInfo extension object, reusing the sending port of the span data. Whether to collect webinfo and the webinfo sampling strategy can be configured in the configuration file. It supports truncation of message data to prevent UDP protocol transmission failures caused by excessively large data packets; it also supports exception judgement of response messages, which can be used for aggregated analysis.

Limitations: 
① Currently, only agent data supports sending using the Thrift protocol, and gRPC is not supported at present; 
② The collection of request URLs and headers, as well as response headers, depends on the client service running on Tomcat, Jetty, or Undertow containers; 
③ The collection of request bodies and response bodies depends on the service using the Spring framework.


wangjiee1:
1

PR #10009, opened 2023-05-31
fix [UnderTow does not collect data] bug

KingCesc:
UnderTow does not collect data, befause 
[com.navercorp.pinpoint.plugin.undertow.interceptor.ConnectorsExecuteRootHandlerInterceptor#handlerPredicate ] return a Predicate， but the Predicate‘s test method always return false.
image


emeroad:
We will review this PR.

jaehong-kim (approved):

PR #10013, opened 2023-06-02
[#noissue] The dynamic control pinpoint(Master switch, sampling rate,…

dongduang:
Dynamic Control Pinpoint
We can dynamically control the following functions through nacos：
1.pinpoint collection main switch;
2.pinpoint sampling rate and its switch;
3.collect address;
4.the amount of data sent by stat;
5.log level.
To use, it is necessary to modify the 'profiler. remote. config. addr' in release/point-env. config, as well as the username, password, and namespace in the DefaultApplicationContext.java and DefaultRemoteConfigMonitor.java


emeroad:
It may take a considerable amount of time to review the issue.
Thank you for your understanding.

PR #10140, opened 2023-07-18
[#noissue] Added Logs to record route request

LoggingResearch:
This PR introduces a new logging statement to the handleRouteRequest() method in the stream handling logic of our application. The new log statement will record the details of each request being handled, as well as the remote address from which the request originated.

By logging this information, we aim to improve the traceability and debuggability of our request handling procedures. This will be particularly beneficial in instances of unexpected behavior or failures, as it will allow us to pinpoint the exact request and source of any issues.


emeroad:

Issue #8929, opened 2022-06-16
Migrate FE framework from Angular to React


binDongKim:
As there's some limitations in using Angular in FE development, we have decided to migrate it to React.

The migration work is currently on its way on the [frontend-v3](https://github.com/pinpoint-apm/pinpoint/tree/frontend-v3) branch.

PR #10189, opened 2023-08-03
[#8929] Provide servermap ref

binDongKim:
- Add programmatic event type

### DIFF
--- a/web-frontend/src/main/v3/packages/server-map/src/ui/ServerMap.tsx
+++ b/web-frontend/src/main/v3/packages/server-map/src/ui/ServerMap.tsx
@@ -10,7 +10,11 @@ import { keyBy } from 'lodash';
 
 cytoscape.use(dagre);
 
-type ClickEventHandler<T> = (param: { data?: T; eventType: 'right' | 'left'; position: cytoscape.Position }) => void;
+type ClickEventHandler<T> = (param: {
+  data?: T;
+  eventType: 'right' | 'left' | 'programmatic';
+  position: Partial<cytoscape.Position>;
+}) => void;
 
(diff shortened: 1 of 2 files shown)


web-frontend/src/main/v3/packages/server-map/src/ui/ServerMap.tsx
```diff
@@ -10,7 +10,7 @@ import { keyBy } from 'lodash';
 
 cytoscape.use(dagre);
 
-type ClickEventHandler<T> = (param: { data?: T; eventType: 'right' | 'left'; position: cytoscape.Position }) => void;
+type ClickEventHandler<T> = (param: { data?: T; eventType: 'right' | 'left' | 'programmatic'; position: cytoscape.Position }) => void;
```
BillionaireDY: type `ClickEventHandler` should probably be like this.  

```ts
type ClickEventHandler<T> = (param: {
  data?: T;
  eventType: 'right' | 'left' | 'progmmatic';
  position: Partial<cytoscape.Position>;
}) => void;

```

Because of this https://github.com/pinpoint-apm/pinpoint/blob/b38493403610c830ccadddc40be19aed8854f584/web-frontend/src/main/v3/packages/server-map/src/ui/ServerMap.tsx#L257-L260

BillionaireDY:
I think a [document](https://github.com/pinpoint-apm/pinpoint-fe-docs) needs to be drawn up

web-frontend/src/main/v3/packages/server-map/src/ui/ServerMap.tsx
binDongKim: Applied review

BillionaireDY (approved):
👍

Issue #6512, opened 2020-02-18
While Tracing Kafka listening trace throwing -  java.lang.NullPointerException: callerApplicationName


dinesh4747:
#### Prerequisites
Please check the [FAQ](http://naver.github.io/pinpoint/faq.html), and search existing [issues](https://github.com/naver/pinpoint/issues) for similar questions before creating a new issue.**YOU MAY DELETE THIS PREREQUISITES SECTION.**

- [x] I have checked the [FAQ](http://naver.github.io/pinpoint/faq.html), and [issues](https://github.com/naver/pinpoint/issues) and found no answer.

#### What version of pinpoint are you using?
V2.0.0-RC2

#### Describe your problem**
Unable to get the call stack for spring boot which listening from the Kafka topic and receiving messages from the topic and It was throwing below exceptions -

1. Failed to handle span=version: 1 transactionId { agentStartTime: 1582004417624 sequence: 5 } spanId: 3340841534782189573 parentSpanId: -1 startTime: 1582007056351 elapsed: 2 apiId: 331 serviceType: 8660 acceptEvent { rpc: "kafka://topic=pharos-error?partition=2&offset=313873" endPoint: "stable.ss-kafka-npe-xxxx:9092" remoteAddr: "stable.ss-kafka-npe-xxxxx:9092" }

2. java.lang.NullPointerException: callerApplicationName

#### What have you done?
Steps you have taken to trigger the problem, for example:
1. Launched '...' with agent
2. Tried tracing '...' (see agent log below)
3. Collector outputs trace log (see collector log below)
4. No trace shows up in the web UI (see web log below)

#### Logs

[...]

PR #10536, opened 2023-12-01
[#6512] flink version upgrade

minwoo-jung:


minwoo-jung:
ref : https://github.com/pinpoint-apm/pinpoint/issues/10537

PR #10601, opened 2024-01-09
Fix mismatched version numbers

inRemark:
There are some version numbers errors on the 2.5.x.

Somebody want use this branch but it  can't build success directly after clone.


minwoo-jung:
good!

Issue #10609, opened 2024-01-15
There is a problem with accessing the WebSite.


nine03:
#### What version of pinpoint are you using?
master

#### Describe the bug
The bug is a Live Demo link error in README.md

#### What did you do to trigger the bug?
Steps to reproduce, for example:
1. I've accessed it with that url(http://125.209.240.10:10123/main/ApiGateway@SPRING_BOOT/5m?inbound=1&outbound=4&wasOnly=false&bidirectional=false), but I can't WebSite.
2. The link to Live Demo (http://125.209.240.10:10123/) could not be reached on the https://pinpoint-apm.github.io/pinpoint/ WebSite either.

#### Screenshots
![1](https://github.com/pinpoint-apm/pinpoint/assets/60682087/f0b36f39-0975-4696-bfbf-6078d9bc0ab3)

![2](https://github.com/pinpoint-apm/pinpoint/assets/60682087/220316c9-781e-42a0-8c73-8aa2a385a390)

PR #10610, opened 2024-01-15
Fixed Change last exclamation point to dot and link issue for browsers

nine03:
#10609 

1. I've accessed it with that url(http://125.209.240.10:10123/main/ApiGateway@SPRING_BOOT/5m?inbound=1&outbound=4&wasOnly=false&bidirectional=false), but I can't WebSite.
2. The link to Live Demo (http://125.209.240.10:10123/) could not be reached on the https://pinpoint-apm.github.io/pinpoint/ WebSite either.

Hello, I'm so curious about why I can't access the site. Can you teach me, please.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ The current stable version is [v2.5.3](https://github.com/pinpoint-apm/pinpoint/
 
 ## Live Demo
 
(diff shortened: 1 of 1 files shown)


ga-ram:
Duplicate issue: https://github.com/pinpoint-apm/pinpoint/issues/10520
Currently demo web is not available.

nine03:
@ga-ram

nine03:
[ga-ram](https://github.com/ga-ram)

I see there is the same issue. Thank you so much for your reply.

PR #10663, opened 2024-02-02
[#NotAssigned] Resolve issue for rewritten launcher on spring-boot 3.2

koo-taejin:
### description 
1. location 
    - spring-boot < 3.2
      - `org.springframework.boot.loader.Launcher `
    - spring-boot >= 3.2
      - `org.springframework.boot.loader.launch.Launcher`

2. method signature 
    - spring-boot < 3.2
      - `protected void launch(String[] args, String mainClass, ClassLoader classLoader) throws Exception`
    - spring-boot >= 3.2
      - `protected void launch(ClassLoader classLoader, String mainClassName, String[] args) throws Exception`


donghun-cho:
Due to the integration of the agent module, the file path has been changed from `plugins/spring-boot` to `agent-module/plugins/spring-boot`.
If you could apply the change or allow me to make the modification, I would like to merge this after change.

koo-taejin:
Feel free to change it. 
I think you already have the permissions, right? 
If there's anything else I need to check, feel free to let me know

Thanks as always. 🙇

PR #10664, opened 2024-02-02
[#NotAssigned] Support for Jetty Using Jakarta Servlet

koo-taejin:
### description 
Support for Jetty Using Jakarta Servlet. 
With this, PINPOINT can track data in Spring Boot 3.x when using Jetty, just as before.

Thank 😄


jaehong-kim:
Thank you.

PR #10798, opened 2024-03-27
Fix `NoSuchMethodError` thrown by `ClientResponseFunctionInterceptor` in reactive plugin.

JooHyukKim:
## Motivation

This PR fixes `NoSuchMethodError` thrown by `ClientResponseFunctionInterceptor`.
To be specific,

```java
java.lang.NoSuchMethodError: `int org.spring....<rest-ommitted>.reactive.ClientHttpResponse.getRawStatusCode()`
```

## Explanation

My application runs on SpringBoot 3 uses 6+ version of `spring-reactive` where there is no `getRawStatusCode()`.

I understand that as per Pinpoint version `2.5.x` README file, it says

```markdown
* Range: org.springframework/spring-webflux [5.0.0.RELEASE, 5.3.max]
```
... saying it does not need to conform to version range out of range.
But since there is already method `getStatusCode()` from which we can achieve the same functionality, there is no reason not to use `getStatusCode()` instead.?

## Note

I simply chained ...

```java
response.getStatusCode().value()
```
... because it's never null.
LMK if we prefer null check.

image

### DIFF
--- a/plugins/spring-webflux/src/main/java/com/navercorp/pinpoint/plugin/spring/webflux/interceptor/ClientResponseFunctionInterceptor.java
+++ b/plugins/spring-webflux/src/main/java/com/navercorp/pinpoint/plugin/spring/webflux/interceptor/ClientResponseFunctionInterceptor.java
@@ -52,7 +52,7 @@ public class ClientResponseFunctionInterceptor extends SpanEventSimpleAroundInte
 
         if (args[0] instanceof ClientHttpResponse) {
             ClientHttpResponse response = (ClientHttpResponse) args[0];
-            recorder.recordAttribute(AnnotationKey.HTTP_STATUS_CODE, response.getStatusCode());
+            recorder.recordAttribute(AnnotationKey.HTTP_STATUS_CODE, response.getStatusCode().value());
             this.responseHeaderRecorder.recordHeader(recorder, response);
         }
     }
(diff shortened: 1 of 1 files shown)


JooHyukKim:
May I ask if 2.5.x version is still in maintained, @emeroad? I am asking this because it seems like only PR's against 3.x version have been merged.

emeroad:
@intr3p1d 
https://github.com/pinpoint-apm/pinpoint/commit/a4a504c8c5a3dfd733dc8c10cb8cc64c3c213790#diff-22feb5f82c89d53ec5b0e346a8502d067c5d448439b86a99a7fb7e79c7f4c867
We will backport the above patch to version 2.5.x.

JooHyukKim:
> @intr3p1d [a4a504c#diff-22feb5f82c89d53ec5b0e346a8502d067c5d448439b86a99a7fb7e79c7f4c867](https://github.com/pinpoint-apm/pinpoint/commit/a4a504c8c5a3dfd733dc8c10cb8cc64c3c213790#diff-22feb5f82c89d53ec5b0e346a8502d067c5d448439b86a99a7fb7e79c7f4c867) We will backport the above patch to version 2.5.x.

Great, I was wondering if we should suggest backporting that commit 🙏🏼🙏🏼 
Looks better than this fix, thank you!

May I ask when 2.5.4 is expected to be released?
If it takes too long, we are considering to build and use latest 2.5.x version with the patch.

Issue #10832, opened 2024-04-02
Backport: Support `getStatusCode()` for compatibility with Spring WebFlux 6.1 & Boot 3.x.


intr3p1d:
#10706
#10736 
#10798

PR #10833, opened 2024-04-02
[#10832] Backport: Support `getStatusCode()` for compatibility with S…

intr3p1d:
…pring WebFlux 6.1 & Boot 3.x.

[#NotAssigned] Removed getRawStatusCode() method for removal in Spring Webflux 6.1

[#10736] Separate dummy java class not to disturb spring-web dependency

[#10736] Remove dummy classes after install the plugin

[#10736] Use maven-clean-plugin instead of exec

[#10736] Split spring-stub into separate modules


JooHyukKim:
May I ask the release date of `2.5.4` version? @intr3p1d

PR #11064, opened 2024-05-27
[#noissue] Fix filteredMap merge logic

binDongKim:
- Pass the previous data if there's no data for the same key(agent) in the merging scatter process
- Match servermap data with timestamp, not with the data index
- Consider unauthorized condition only when there is node-list


BillionaireDY:
LGTM 👍

BillionaireDY (approved):

PR #11223, opened 2024-07-04
[#noissue] Support colon-base time format

binDongKim:


web-frontend/src/main/v3/packages/constants/src/date.ts
```diff
@@ -1,4 +1,8 @@
 export const SEARCH_PARAMETER_DATE_FORMAT = 'yyyy-MM-dd-HH-mm-ss';
+export const SEARCH_PARAMETER_DATE_FORMAT_WHITE_LIST = [
+  'yyyy-MM-dd-HH-mm-ss',
```
BillionaireDY: How about using `SEARCH_PARAMETER_DATE_FORMAT ` instead of string

BillionaireDY (approved):
LGTM

PR #11229, opened 2024-07-09
[#noissue] Feat global search

BillionaireDY:
- support 
  - txid search
  - navigate pages

image


binDongKim:
## Check point
- Search should cover case-insensitive-ly 
  - ex) Search `servermap` doesn't work
- Rename `useService` hook as `useMenu` or `useMenuItem`
  - Locate the file under the `apps/web/src/hooks` path
- Rollback `path.ts` file location if it's not necessary

PR #11303, opened 2024-07-30
[#10776] Implement micrometer ui

binDongKim:
- Base work
  - routing, sidebar, ...
  - not open it yet (only access through url)
- Metric definition form
- Chart component from shadcn/ui


BillionaireDY (approved):
LGTM

PR #11535, opened 2024-09-29
[#noissue][FE] fix: filtermap scatter to transactionList bug fix

BillionaireDY:
Fixes bugs that occur when dragging a scatter chart on the filterMap page.


web-frontend/src/main/v3/packages/constants/src/locales/en.json
```diff
@@ -52,7 +52,8 @@
   },
   "TRANSACTION_LIST": {
     "SELECT_TRANSACTION": "Select your transaction",
-    "TRANSACTION_RETRIEVE_ERROR": "Transaction information is missing<br>Will go to Main"
+    "TRANSACTION_RETRIEVE_ERROR": "Transaction information is missing.",
+    "TRANSACTION_RETRIEVE_ERROR_DESC": "Currently, there is no data available from the parent tab. Please try again from the parent tab. (If you close this dialog, you will be redirected to the main page.)"
```
binDongKim: This message will ever be shown to the user only when they try getting to the transaction-list page via url without dragging the scatter chart from the server-map page. So the meaning of the word *parent* tab might not be clear to the user. How about suggesting them to start it over from the server-map page?

```suggestion
    "TRANSACTION_RETRIEVE_ERROR_DESC": "Please start it over from the server-map page. (If you close this dialog, you will be redirected to the server-map page.)"
```

binDongKim (approved):

PR #11885, opened 2025-01-01
Fix compilation - execution still fails

DaGeRe:
This fixes the compilation of the quickstart app, but doesn't fix the execution (yet).


emeroad:

Issue #12025, opened 2025-02-06
Experimental V4 Id feature


emeroad:
Add experimental v4 id feature to Agent for Pinpoint V4

PR #12144, opened 2025-03-11
[#12025] Refactor applicationId to applicationName

emeroad:


minwoo-jung (approved):
There don't seem to be any issues with the changes.

Issue #12623, opened 2025-06-30
PostgreSQL Positional Parameters


feelform:
PostgresSQL positional Parameter
ref: https://github.com/pinpoint-apm/pinpoint-node-agent/issues/336

PostgreSQL Positional Parameters: https://www.postgresql.org/docs/current/sql-expressions.html#SQL-EXPRESSIONS-PARAMETERS-POSITIONAL

```
    @Test
    public void postSqlPositionalParameterTest() {
        DefaultSqlNormalizer dut = new DefaultSqlNormalizer();
        assertEquals("SELECT * FROM member WHERE id = $1", dut.normalizeSql("SELECT * FROM member WHERE id = $1").getNormalizedSql());
    }

expected: <SELECT * FROM member WHERE id = $1> but was: <SELECT * FROM member WHERE id = $0#>
```

Support for PostgreSQL positional parameters for `$1, $2`

PR #12624, opened 2025-06-30
[#12623] PostgreSQL Positional Parameters

feelform:


commons-profiler/src/test/java/com/navercorp/pinpoint/common/profiler/sql/DefaultSqlNormalizerTest.java
```diff
@@ -410,4 +411,21 @@ private void assertEqualObject(String expected) {
...
+            dut.normalizeSql("SELECT * FROM member WHERE id = $122309").getNormalizedSql());
+
+        // Edge case: $ followed by non-digit (e.g., variable-like)
+        assertEquals("SELECT * FROM config WHERE key = $value",
+                dut.normalizeSql("SELECT * FROM config WHERE key = $value").getNormalizedSql());
+
+        // Edge case: $ inside string literal
+        assertEquals("SELECT * FROM logs WHERE message LIKE '0$'",
+                dut.normalizeSql("SELECT * FROM logs WHERE message LIKE '0$'").getNormalizedSql());
+    }
```
kojandy: This test case wasn't working properly because there were no parameters extracted and `parameter.touch()` was not called, which made the `SqlNormalizer.normalizeSql()` reuse the original SQL string.

https://github.com/pinpoint-apm/pinpoint/blob/88cf67a30f1f9fc8f8dac9c3b07ed129f215c313/commons-profiler/src/main/java/com/navercorp/pinpoint/common/profiler/sql/ParserContext.java#L158-L166

To avoid reusing the original query string, at least one parameter should be extracted like this.

```java
@Test
public void testPostSqlPositionalParameter() {
    assertEqual("$123, 123", "$123, 0#");
    assertEqual("$value, 123", "$value, 0#");
    assertEqual("'$123', 123", "'0$', 1#");
    assertEqual("$; 123", "$; 0#");
    assertEqual("$(123); 123", "$(0#); 1#");
    assertEqual("'$''123'", "'0$'");
}
```

commons-profiler/src/main/java/com/navercorp/pinpoint/common/profiler/sql/ParserContext.java
```diff
@@ -140,6 +140,15 @@ public NormalizedSql parse() {
...
                     break;
 
+                case '$':
+                    final int nextChar = lookAhead1(i);
+                    if (nextChar >= '0' && nextChar <= '9') {
+                        numberTokenStartEnable = false;
+                        normalized.append(nextChar);
+                        break;
+                    }
+                    // Fall through for non-positional '$' to be treated other rules.
```
kojandy:
```suggestion
                case '$':
                    final int nextChar = lookAhead1(i);
                    if (nextChar >= '0' && nextChar <= '9') {
                        numberTokenStartEnable = false;
                    }
                    normalized.append(ch);
                    break;
```
It can be implemented by appending the current character `$` and not treating the following number token as a number. With the current implementation, `SELECT * FROM member WHERE id = $1` becomes `SELECT * FROM member WHERE id = 49` because it skips current character and outputs the character '1' as its ASCII value 49.

feelform: `default:` code was changed by emeroad. I have applied your suggestion.
```suggestion
                case '$':
                    final int nextChar = lookAhead1(i);
                    if (nextChar >= '0' && nextChar <= '9') {
                        numberTokenStartEnable = false;
                        normalized.append(nextChar);
                        break;
                    }
                    // Fall through for non-positional '$' to be treated other rules.
```

commons-profiler/src/test/java/com/navercorp/pinpoint/common/profiler/sql/DefaultSqlNormalizerTest.java
feelform: @kojandy 
```
Assertions.assertEquals(actual, normalizedSql, "normalizedSql check");
```
actual argument use expect argument into Assertions.assertEquals in assertEqual. I think It needs to rename refactoring.

kojandy (approved):

commons-profiler/src/test/java/com/navercorp/pinpoint/common/profiler/sql/DefaultSqlNormalizerTest.java
kojandy: I was also confused by these parameters at first too, but changing the name like this is definitely better. 👍

Issue #12760, opened 2025-08-06
ServerMap > Change color variation


jihea-park:
<img width="622" height="318" alt="Image" src="https://github.com/user-attachments/assets/a4985522-7a5b-4f96-b5ab-6210ed95f212" />

PR #12761, opened 2025-08-06
[#12760] ServerMap > Change color variation

jihea-park:


jihea-park:
The color transition can cause significant changes to the individual's perception, so we take a conservative approach and cancel the transition.

Issue #12883, opened 2025-09-10
Update agent unmodifiableFilter of transform


jaehong-kim:
Update to filter classes under the jdk package.

PR #12907, opened 2025-09-16
[#12883] Update agent unmodifiableFilter of transform

jaehong-kim:
#12883


Dongnyoung:
@jaehong-kim 
I added a small test-only PR to cover the newly added `jdk/*` branch,
which fixes the Codecov patch coverage failure.

PR: #13398

PR #13281, opened 2026-01-13
[#noissue] Lower log level for missing classes in ASMClassWriter

Dongnyoung:
Hello, 
### Background
While testing Pinpoint in a real Spring Boot application, I observed that
`ASMClassWriter` logs WARN messages when optional classes are missing.

### Problem
This is a normal fallback path, but logging it as WARN produces noisy logs
and may mislead users into thinking there is a problem.

### Change
Lower the log level from WARN to DEBUG without altering behavior.

### Verification
Reproduced by attaching the Pinpoint agent to a Spring Boot application.

스크린샷 2026-01-14 013546


Dongnyoung:
This change only adjusts the log level (warn → debug) and does not affect runtime behavior.
The missing coverage comes from a null-check branch that is not exercised by existing tests.
Please let me know if you’d prefer a different approach or additional tests.

jaehong-kim:
@Dongnyoung 
Thank you.

PR #14018, opened 2026-07-16
[#14016] Fix apdex and avgTimeMs returning 0 in uriStat summary

### LINKED ISSUE #14016: URI stat summary returns apdex/avgTimeMs as 0 when type parameter is omitted
`GET /api/uriStat/summary` (without the `type` parameter) always returns `apdex: 0.0` and `avgTimeMs: 0.0` regardless of the actual data. Calls with `type=...` return correct values.

### Cause
#11944 (65091311dc) renamed `UriStatSummaryEntity` fields `apdexRaw` → `totalApdexRaw` and `totalTimeMs` → `sumOfTotalTimeMs`, and updated the aliases in the shared `selectStatSummary` SQL accordingly — but the plain `uriStatSummary` query in `UriStatSummaryMapper.xml` kept the old aliases:

```sql
sum(apdexRaw) as apdexRaw,
sum(totalTimeMs) as totalTimeMs,
```

The result map relies on column-name auto-mapping, so these two columns silently no longer match any entity field. `entity.totalApdexRaw` / `entity.sumOfTotalTimeMs` stay 0, and `toApdex()` / `toAvgTimeMs()` (MapperUtils) compute `0 / totalCount = 0`.

Other fields (`totalCount`, `maxTimeMs`, `failureCount`) match and are correct, so the response looks plausible — no error is raised anywhere.

### Fix
Rename the two aliases in the `uriStatSummary` query to match the entity fields: `as totalApdexRaw`, `as sumOfTotalTimeMs`.

minwoo-jung:
resolves #14016

### Cause
#11944 renamed `UriStatSummaryEntity` fields `apdexRaw` → `totalApdexRaw` and `totalTimeMs` → `sumOfTotalTimeMs` and updated the aliases in the shared `selectStatSummary` SQL, but the plain `uriStatSummary` query kept the old aliases. Since the result map relies on column-name auto-mapping, those two columns silently stopped matching any entity field, so `GET /api/uriStat/summary` (without `type`) always returned `apdex: 0.0` and `avgTimeMs: 0.0`.

### Change
Rename the two aliases in the `uriStatSummary` query to match the entity fields. 2-line mapper XML change, no Java change.

### DIFF
--- a/uristat/uristat-web/src/main/resources/mapper/uristat/UriStatSummaryMapper.xml
+++ b/uristat/uristat-web/src/main/resources/mapper/uristat/UriStatSummaryMapper.xml
@@ -7,10 +7,10 @@
     <select id="uriStatSummary" resultMap="uriStatSummaryEntity" parameterType="UriStatSummaryQueryParameter">
         SELECT
         uri,
-        sum(apdexRaw) as apdexRaw,
+        sum(apdexRaw) as totalApdexRaw,
         sum("count") as totalCount,
         max(maxLatencyMs) as maxTimeMs,
-        sum(totalTimeMs) as totalTimeMs,
+        sum(totalTimeMs) as sumOfTotalTimeMs,
         sum(failureCount) as failureCount,
         version
         FROM uriStat


intr3p1d:
LGTM

Issue #14044, opened 2026-07-22
Kotlin coroutine interceptor can leave an async trace block open


cj848:
## Bug description

`ResumeWithInterceptor` can leave the coroutine span event opened on a `ChildTrace` when another async integration changes the `AsyncContext`'s current trace while `BaseContinuationImpl.resumeWith(...)` is executing.

`AsyncContextSpanEventSimpleAroundInterceptor.before()` obtains a trace with `continueAsyncTraceObject(...)` and starts a span event. Its `after()` resolves the trace again through `currentAsyncTraceObject()`. If that lookup returns `null` or a different trace after a nested Reactor callback, the original event is not closed. When Reactor later closes the child trace's asynchronous invocation block, Pinpoint reports a mismatched stack ID because the coroutine event is still on top.

## Observed warning

```text
Corrupted call stack found
PinpointException: not matched stack id. expected=1001, current=-1,
spanEvent=SpanEvent{stackId=-1, serviceType=8901, ...}
PinpointException: not empty call stack
PinpointException: already closed trace
```

The application request succeeds, but its trace can be corrupted or dropped and the agent emits repeated warning stack traces.

## Environment

- Pinpoint agent: 3.1.0
- Spring WebFlux / Reactor Core: 3.8.6
- Kotlin coroutines: 1.11.0
- JDK: 26
- Coroutine execution uses a virtual-thread dispatcher

## Expected behavior

The interceptor should close the same trace block that it opened in `before()`, regardless of changes to the async context's current trace during the intercepted continuation.

## Proposed fix

Use the block-aware async interceptor base for `ResumeWithInterceptor`. `AsyncContextSpanEventBlockSimpleAroundInterceptor` passes the exact `TraceBlock` returned by `before()` into `after()`, avoiding a second lookup of `currentAsyncTraceObject()`.

I am preparing a pull request with focused regression coverage.

jaehong-kim:
Great catch and a clean fix

PR #14047, opened 2026-07-22
[#14044] Strengthen coroutine lifecycle regression

cj848:
## Why

Follow-up to #14046. The original unit test verifies the block-aware callback contract, but reverting the production change fails at the interface assertion before exercising the stale-current-trace behavior.

This revision keeps both callback contracts executable in the test and models the lifecycle observed around coroutine/Reactor nesting: `before()` starts trace A, a simulated nested callback clears the async context's current trace, and `after()` must still close the exact block from trace A.

Related to #14044.

## What

- Simulate the current async trace being cleared between `ResumeWithInterceptor.before()` and `after()`.
- Exercise either the pre-fix `AroundInterceptor` or fixed `BlockAroundInterceptor` callback contract from the same test source.
- Verify no second `currentAsyncTraceObject()` lookup and preserve exact block identity plus enter/begin/leave/close ordering.

## Regression proof

- Pre-fix superclass: `ResumeWithInterceptorTest` fails because `AsyncContextSpanEventSimpleAroundInterceptor.after()` calls `currentAsyncTraceObject()` after it has been cleared.
- Block-aware superclass: the same test passes and closes the block returned by `before()`.

## Verification

- `./mvnw -q -pl agent-module/plugins/kotlin-coroutines -am -DskipITs -Dtest=ResumeWithInterceptorTest -Dsurefire.failIfNoSpecifiedTests=false test`
- `./mvnw -q -pl agent-module/plugins/kotlin-coroutines -am -DskipITs test`
- `Coroutines_1_6_IT`: 38 tests, 0 failures/errors/skips
- `git diff --check`


cj848:
Closing this follow-up to keep the upstream change represented by the already merged #14046. The behavioral test refinement will not be proposed as a separate PR.

Issue #14039, opened 2026-07-21
Query serviceMap by service unit


minwoo-jung:
Enhance `GET /api/servermap/serviceMap` to query the application map by service unit.

- Resolve `serviceName` from the request (`@ServiceParam`) and pass the resolved service uid to the map query
- When the service is not DEFAULT: ignore `applicationName` and build the map from all applications belonging to the service (multi-seed `MapServiceOption`)
- When the service is DEFAULT: keep the existing single-application behavior
- The requested service is always rendered expanded; the `keepServiceNames` parameter is removed
- Search options are fixed to defaults (depth 1/1, unidirectional); `SearchOptionForm` parameters are removed from this API
- A service with no applications returns an empty map instead of an error

minwoo-jung:
<img width="4252" height="1614" alt="Image" src="https://github.com/user-attachments/assets/75aeb613-4ffa-44a1-800c-828c192b6e29" />

minwoo-jung:
url format example  :
http://localhost:8080/api/servermap/serviceMap?applicationName=test_minwoo_app&serviceTypeName=SPRING_BOOT&from=20260721T011337Z&to=20260721T071337Z&useStatisticsAgentState=true

PR #14056, opened 2026-07-23
[#14039] Remove obsolete query parameters from serviceMap request

minwoo-jung:
## Summary
- #14040 dropped SearchOption (callerRange/calleeRange/bidirectional/wasOnly) and keepServiceNames from `GET /serviceMap`; the service map is now queried per service unit with a fixed search depth.
- Stop sending these now-ignored parameters from the FE (`GetServiceMap.Parameters`, `useGetServiceMap`, `ServiceMapFetcher`).

## Test plan
- [ ] `/serviceMap` request URL no longer includes callerRange/calleeRange/bidirectional/wasOnly/keepServiceNames
- [ ] Service map renders correctly with the remaining parameters (no regression)


jihea-park (approved):
👍

PR #14082, opened 2026-07-29
[#noissue] Narrow the agent management error boundary to the table area

jihea-park:
## Summary
- The Agent management page wrapped its entire content in a single `ErrorBoundary`, so a failed `/api/agents` request replaced the header and the application selector too — leaving the user with no way to pick another application and recover.
- Split the data-fetching part out into `AgentManagementContent` and wrapped only that in the `ErrorBoundary` + `Suspense`. The title and application selector now stay mounted while the table area shows the error fallback.
- Added the selected application to the boundary's `resetKeys`, so choosing a different application clears the error state and refetches automatically.

## Test plan
- [ ] Open Configuration > Agent management and select an application — the agent table loads as before.
- [ ] Force `/api/agents` to fail (e.g. block it in devtools) — only the table area shows the error fallback; the title and application selector remain usable.
- [ ] While the error fallback is shown, select a different application — the error clears and the list is refetched.
- [ ] Remove an agent and remove an application — toasts and list refresh behave as before, and removing an application still clears the selection.
- [ ] `yarn build`, `yarn test`, `yarn lint`

### DIFF
--- a/web-frontend/src/main/v3/packages/ui/src/components/Config/agentManagement/AgentManagementFetcher.tsx
+++ b/web-frontend/src/main/v3/packages/ui/src/components/Config/agentManagement/AgentManagementFetcher.tsx
@@ -1,8 +1,7 @@
 import React from 'react';
 import { useTranslation } from 'react-i18next';
-import { Configuration, ApplicationType, AgentManagementList } from '@pinpoint-fe/ui/src/constants';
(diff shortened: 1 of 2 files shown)


jihea-park:
### as-is
스크린샷 2026-07-29 오후 3 17 03

### to-be
스크린샷 2026-07-29 오후 3 13 54

PR #14090, opened 2026-08-02
[#noissue] Add Ktor client tracing instrumentation

cj848:
## Summary

Add outbound-call tracing for Ktor HTTP client (`io.ktor.client.plugins.HttpSend$DefaultSender.execute(...)`) to the built-in ktor plugin.

- Each client call emits a `KTOR_CLIENT` span event with propagation headers.
- Each suspending call also emits a `KTOR_CLIENT_INTERNAL` async span event, completed when the suspending continuation resumes.
- Retry attempts via `HttpRequestRetry` emit one span event per attempt (each retry is a distinct remote call).
- Mirrors a custom plugin maintained outside this repo in production deployments. Outbound client tracing is enabled by default (since the review change); set `-Dprofiler.ktor.client.enable=false` to opt out.

## Needs review

- ~~**Service type codes**: claims `KTOR_CLIENT=1900` / `KTOR_CLIENT_INTERNAL=1901`.~~ → **Resolved per review**: `9068`/`9069` in the RPC range (9067 is taken by `JDK_HTTPURLCONNECTOR_INTERNAL` via Java SPI; 9060/9070 are RESERVED). Registered in the `ServiceType.java` javadoc table.
- **Test coverage**: 28 unit tests cover interceptors, adaptors, configuration, and scope semantics. Full agent-anchored integration (drive the jar through a real `naver-pinpoint-agent` tarball plus a suspended `HttpRequestRetry` continuation) is possible but needs the upstream IT harness; happy to add as a follow-up PR if desired.
- **Dependency on the Kotlin-coroutines plugin**: same constraint as the existing ktor server plugin's SuspendFunctionGun pattern; let me know if you want it removed.

## Open questions

1. ~~Should `profiler.ktor.client.enable` default flip to true once this lands?~~ → **Yes (review)** — flipped, and the client keys are now in the shipped release/local `pinpoint.config`.
2. ~~Want the Ktor client version compat range documented in README (validated against Ktor 3.5.1 clients in production)?~~ → **Yes (review)** — documented as `io.ktor/ktor-client-core-jvm [2.3, 3.5]`, backed by artifact-level `javap` checks (2.3.12 / 3.0.0 / 3.5.1 / 3.5.2 all structurally identical at the instrumentation targets).

Refs: https://pinpoint-apm.github.io/pinpoint/plugindevguide.html

### DIFF
--- a/agent-module/agent/src/main/resources/profiles/local/pinpoint.config
+++ b/agent-module/agent/src/main/resources/profiles/local/pinpoint.config
@@ -583,6 +583,15 @@ profiler.ktor.server.excludeurl=
 # Retransform
 profiler.ktor.http.server.retransform.configure-routing=true
 
+# Client (outbound)
+# Traces io.ktor.client.plugins.HttpSend$DefaultSender.execute(...) and records
+# an outbound span event per HTTP attempt (including HttpRequestRetry retries).
+profiler.ktor.client.enable=true
+# record the request URL on the span event
+profiler.ktor.client.param=true
+# treat a thrown Throwable from the client as an error on the span event
+profiler.ktor.client.mark.error=true
+
 ###########################################################
 # JSP                                                     #
 ###########################################################
--- a/agent-module/agent/src/main/resources/profiles/release/pinpoint.config
+++ b/agent-module/agent/src/main/resources/profiles/release/pinpoint.config
@@ -579,6 +579,15 @@ profiler.ktor.server.excludeurl=
 # Retransform
 profiler.ktor.http.server.retransform.configure-routing=true
 
+# Client (outbound)
+# Traces io.ktor.client.plugins.HttpSend$DefaultSender.execute(...) and records
+# an outbound span event per HTTP attempt (including HttpRequestRetry retries).
+profiler.ktor.client.enable=true
+# record the request URL on the span event
+profiler.ktor.client.param=true
+# treat a thrown Throwable from the client as an error on the span event
+profiler.ktor.client.mark.error=true
+
 ###########################################################
 # JSP                                                     #
 ###########################################################
--- a/agent-module/plugins/ktor/README.md
+++ b/agent-module/plugins/ktor/README.md
@@ -35,4 +35,14 @@ profiler.ktor.server.excludeurl=
 # Retransform
 profiler.ktor.http.server.retransform.configure-routing=true
 
+# Client (outbound)
+# Traces io.ktor.client.plugins.HttpSend$DefaultSender.execute(...) and records
+# an outbound span event per HTTP attempt (including HttpRequestRetry retries).
+# Disabled by default — opt in only when the JVM runs a Ktor HTTP client.
+profiler.ktor.client.enable=false
+# record the request URL on the span event, default true
+profiler.ktor.client.param=true
+# treat a thrown Throwable from the client as an error on the span event, default true
+profiler.ktor.client.mark.error=true
+
 ~~~
--- a/agent-module/plugins/ktor/README.md
+++ b/agent-module/plugins/ktor/README.md
@@ -1,6 +1,9 @@
 ## Ktor
 * Since: Pinpoint 3.0.1
 * See: https://ktor.io/
+* Range (client): io.ktor/ktor-client-core-jvm [2.3, 3.5]
+  (instrumentation targets `HttpSend$DefaultSender.execute` and its generated
+   continuation verified structurally identical in the 2.3.12, 3.0.0, 3.5.1 and 3.5.2 artifacts)
 
 ### Pinpoint Configuration
 pinpoint.config
@@ -34,5 +37,15 @@ profiler.ktor.server.excludeurl=
 
 # Retransform
 profiler.ktor.http.server.retransform.configure-routing=true
+
+# Client (outbound)
+# Traces io.ktor.client.plugins.HttpSend$DefaultSender.execute(...) and records
+# an outbound span event per HTTP attempt (including HttpRequestRetry retries).
+# Enabled by default; set false to skip the Ktor client transforms.
+profiler.ktor.client.enable=true
+# record the request URL on the span event, default true
+profiler.ktor.client.param=true
+# treat a thrown Throwable from the client as an error on the span event, default true
+profiler.ktor.client.mark.error=true
 
 ~~~
--- a/agent-module/plugins/ktor/pom.xml
+++ b/agent-module/plugins/ktor/pom.xml
@@ -40,6 +40,18 @@
             <version>2.3.12</version>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>io.ktor</groupId>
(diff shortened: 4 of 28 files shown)


cj848:
Hi @jaehong-kim, hope you're doing well.

This PR has been waiting for a review for about two weeks, so I wanted to check in politely. When you have time, could you take a look? Your input on the two points in the *Needs review* / *Open questions* sections of the PR body would be especially valuable: the `KTOR_CLIENT` / `KTOR_CLIENT_INTERNAL` service type code slots, and whether depending on the Kotlin-coroutines plugin is acceptable for the built-in ktor plugin.

If you'd rather discuss this in an issue first (as we did for the coroutine trace fix in #14044), I'm happy to open one and rework from there. Thank you!

jaehong-kim:
Thanks for the contribution.

1. Move KTOR_CLIENT / KTOR_CLIENT_INTERNAL out of the SERVER code range
Per ServiceTypeCategory, 1000–1999 is the SERVER range.
Please refer to the javadoc of com.navercorp.pinpoint.common.trace.ServiceType

Open questions
1. Default value — Yes
2. Version compatibility — Yes, please document it in the README

cj848:
Hi @jaehong-kim,

All three points addressed in `9936c4ebe`.

**1. Moved `KTOR_CLIENT` / `KTOR_CLIENT_INTERNAL` out of the SERVER range → `9068` / `9069` (RPC range)**

Correct, thanks — `ServiceTypeCategory.SERVER(1000, 1999)` covers my original `1900`/`1901` slots. They now sit in the RPC range next to the other HTTP client plugins, in the pair order used by every other client (CLIENT then _INTERNAL).

Two things worth recording for slot safety:
- `9067` is not actually free: `jdk-http` registers `JDK_HTTPURLCONNECTOR_INTERNAL` at `9067` via Java SPI (`ServiceTypeFactory.of(9067, ...)` in `JdkHttpConstants.java`, no type-provider.yml), and `TraceMetadataLoader` throws `RuntimeException` at agent start on duplicate ServiceType codes — both plugins ship in the same `plugins/assembly/pom.xml`, so a 9067 pick would have broken every agent deployment. I caught this by enumerating registrations across all three registration paths (every `type-provider.yml`, every plugin `ServiceTypeFactory.of(...)`, commons) instead of only the YAML files.
- Full-slot picture upstream of my change: `9050`–`9059` HTTP clients, `9060` RESERVED, `9062`/`9063` httpclient5, `9065`/`9066` jdk-httpclient, `9067` jdk-http, `9070` RESERVED. So `9068`/`9069` is the next free adjacent pair. Also registered both rows in the `ServiceType.java` javadoc table (italic, same style as httpclient5 / jdk-httpclient).

**Open question 1 — default value: flipped to `true`.** `KtorPluginConfig` now defaults `profiler.ktor.client.enable` to `Boolean.TRUE`, and the three client keys (`enable`, `param`, `mark.error`) are added to the shipped `profiles/release/pinpoint.config` and `profiles/local/pinpoint.config` — previously they existed only in the plugin README. The shipped release profile keeps `profiler.ktor.enable=false`, so nothing new turns on unless the user enables the ktor plugin for their app.

**Open question 2 — version compatibility: documented.** The README now carries:
`Range (client): io.ktor/ktor-client-core-jvm [2.3, 3.5]` — built against 2.3.12, validated against Ktor 3.5.1 clients in production deployments.

**Verification:** `./mvnw -q -pl agent-module/plugins/ktor clean test` → 28 tests, 0 failures (root pom sets `testFailureIgnore=true`, so the count is from the surefire XML, not the exit code). Two new config tests cover the flip (`clientEnabledByDefault`, `clientDisabledFlagRespected`).

One pre-existing note I noticed while moving the entries (left as-is since it's not part of your three points): `KTOR_CLIENT_INTERNAL`'s matcher targets annotation key 46 (`http.status.code`, from 524bf27ad), while the other HTTP-client internals — httpclient5 `9063`, jdk-httpclient `9066`, jdk-http `9067` — all match 48 (`http.internal.display`). Today `TraceHolder` records no attribute annotation on the internal span event at all, so the display argument resolves blank either way; the main span event records `http.url`(40) via `ClientRequestRecorder`. If you'd like consistency, I can switch the matcher to 48 (and optionally record an internal-display string) as a follow-up — otherwise happy to keep 46 if it's intentional.

cj848:
Follow-up on the version-compat note: I replaced the production-anecdote wording with a verifiable check (README updated in `dccaa9c31`).

**Check**: the agent's client instrumentation only requires these two internal shapes to exist, matched by name at runtime:

1. `io.ktor.client.plugins.HttpSend$DefaultSender.execute(HttpRequestBuilder, Continuation)`
2. `io.ktor.client.plugins.HttpSend$DefaultSender$execute$1` — constructor `(DefaultSender, Continuation)` plus `invokeSuspend(Object)`

**Result** — identical signatures in every artifact fetched from Maven Central:

| version | `execute` | `$execute$1` ctor | `invokeSuspend` |
|---|---|---|---|
| 2.3.12 | ✓ | ✓ | ✓ |
| 3.0.0 | ✓ | ✓ | ✓ |
| 3.5.1 | ✓ | ✓ | ✓ |
| 3.5.2 (latest) | ✓ | ✓ | ✓ |

**Reproduce** (any version `V`):
```bash
curl -sLO https://repo1.maven.org/maven2/io/ktor/ktor-client-core-jvm/V/ktor-client-core-jvm-V.jar
javap -cp ktor-client-core-jvm-V.jar 'io.ktor.client.plugins.HttpSend$DefaultSender'
javap -p -cp ktor-client-core-jvm-V.jar 'io.ktor.client.plugins.HttpSend$DefaultSender$execute$1'
```

Caveat unchanged: structural sameness only guarantees the transformers *match*; end-to-end tracing behavior on a given ktor runtime still needs the agent-anchored integration harness proposed in *Needs review*."

(note: "compile against the lowest version proves compatibility" does not hold for bytecode-instrumentation plugins — compilation proves the plugin's own code builds; the targets here are compiler-internal (`$execute$1`) and aren't covered by any semver guarantee, hence the per-artifact check above.)

jaehong-kim:
Thanks for the quick turnaround on the review items — the updates all look good.

Could you squash the 8 commits into one before merge? 
It's a team convention here — one PR, one commit — and please use the repo's commit message format 
(e.g. [#noissue] Add Ktor client tracing instrumentation)

cj848:
@jaehong-kim Done — the 8 commits are squashed into a single commit and the PR title now follows the same convention:

- commit: `499bd5e6a [#noissue] Add Ktor client tracing instrumentation`
- PR title: `[#noissue] Add Ktor client tracing instrumentation`

The tree is byte-identical to the reviewed head (`dccaa9c31`), so the approved diff is unchanged; only the history was rewritten (force-pushed with lease). SonarCloud checks on the new head already passed.

(I tried to formally re-request your review but my account lacks the permission on this repo — hence the mention.)

jaehong-kim:
Excellent work — thank you.

PR #14135, opened 2026-08-11
[#14127] return empty PAnnotationValue for NullAnnotation to prevent NPE

### LINKED ISSUE #14127: NullPointerException caused by AnnotationValueMapper
#### Prerequisites
Please check the [FAQ](https://pinpoint-apm.gitbook.io/pinpoint/faq), and search existing [issues](https://github.com/pinpoint-apm/pinpoint/issues) for similar questions before creating a new issue.**YOU MAY DELETE THIS PREREQUISITES SECTION.**

- [√] I have checked the [FAQ](https://pinpoint-apm.gitbook.io/pinpoint/faq), and [issues](https://github.com/pinpoint-apm/pinpoint/issues) and found no answer.

#### What version of pinpoint are you using?
v3.0.3

#### Describe the bug
In Annotations.java, NullAnnotation is created when the annotation value is null. AnnotationValueMapper converts NullAnnotation to null and returns it. However, in PAnnotation.java, if the value is null, an NPE is thrown.

#### What did you do to trigger the bug?
1. Trigger path of NPE
SpanGrpcDataSender.dispatcher.onDispatch() // messageConverter.toMessage(data)
  → GrpcSpanMessageConverter.buildPSpan() //  mapper.map(span,..., pSpan)
    → SpanMessageMapperImpl.map(Span) // Traverse annotations and call map(annotation).
      → SpanMessageMapperImpl.map(Annotation<?>) //  pAnnotation.setValue(annotationValueMapper.map(annotation))
        → AnnotationValueMapper.map() // Return null when NullAnnotation is used.
          → PAnnotation.Builder.setValue(null) // ← NullPointerException is thrown.

When using ES in the code, the version information of ES could not be collected, triggering this bug.

#### Expected behavior
The design of the annotation has issues. Since the NullAnnotation class is designed, when an annotation is null, a runtime exception should not occur during the subsequent generation of Spans.

#### Additional context
The NPE has a great impact and causes span data loss. I have temporarily changed the return null in AnnotationValueMapper to return getAnnotationBuilder().build() as the solution. However, I think this is a design issue and there may be a better solution.

waterWang:
Fixes #14127

## Problem

`AnnotationValueMapper.map(Annotation<?>)` returns `null` when the annotation is a `NullAnnotation`. The caller (`SpanMessageMapper.map(Annotation<?>)` → `PAnnotation.Builder.setValue(...)`) does not guard against a null return value, causing a `NullPointerException`:

```
PAnnotation.Builder.setValue(null)  ← NullPointerException
```

This causes span data loss, for example when ES version information cannot be collected.

## Root cause

The `NullAnnotation` class was designed to represent annotations with a null value, but the mapper's `NullAnnotation` branch returns `null` instead of a valid empty `PAnnotationValue`. The calling code in `SpanMessageMapper` unconditionally calls `pAnnotation.setValue(annotationValueMapper.map(annotation))` without a null check.

## Fix

Return an empty `PAnnotationValue` (via `getAnnotationBuilder().build()`) instead of `null` when the annotation is a `NullAnnotation`. This preserves the semantics of "no annotation value" while preventing the NPE in the caller.

```diff
 default PAnnotationValue map(Annotation<?> annotation) {
     if (annotation instanceof NullAnnotation) {
-        return null;
+        return getAnnotationBuilder().build();
     }
     return mapNonNull(annotation);
 }
```

## Impact

The NPE has a high impact — it causes span data loss. The fix is minimal and safe: returning an empty `PAnnotationValue` is equivalent to the previous behavior of returning `null` from the caller's perspective (the builder accepts an empty value), but avoids the crash.

### DIFF
--- a/agent-module/profiler/src/main/java/com/navercorp/pinpoint/profiler/context/grpc/mapper/AnnotationValueMapper.java
+++ b/agent-module/profiler/src/main/java/com/navercorp/pinpoint/profiler/context/grpc/mapper/AnnotationValueMapper.java
@@ -139,7 +139,9 @@ public interface AnnotationValueMapper {
     @ToPAnnotationValue
     default PAnnotationValue map(Annotation<?> annotation) {
         if (annotation instanceof NullAnnotation) {
-            return null;
+            // Return an empty PAnnotationValue instead of null to avoid NPE in callers
+            // that unconditionally set the result on a PAnnotation.Builder.
+            return PAnnotationValue.getDefaultInstance();
         }
         return mapNonNull(annotation);
     }
--- a/agent-module/profiler/src/test/java/com/navercorp/pinpoint/profiler/context/grpc/mapper/AnnotationValueMapperTest.java
+++ b/agent-module/profiler/src/test/java/com/navercorp/pinpoint/profiler/context/grpc/mapper/AnnotationValueMapperTest.java
@@ -21,7 +21,6 @@ import org.junit.jupiter.api.Test;
 import java.util.Objects;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertNull;
 
 /**
  * @author Woonduk Kang(emeroad)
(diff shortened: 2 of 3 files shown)


emeroad:

PR #14182, opened 2026-08-24
[#noissue] Add unit tests for Ktor client tracing instrumentation

cj848:
## Summary

Supplements unit test coverage for the Ktor client tracing instrumentation merged in 24f56ef (#14090). I missed verifying the test coverage before merging that PR — codecov reported 34.6% patch coverage there — so this PR adds the missing tests.

**Why the coverage was low**: the three largest new classes had no unit tests at all, accounting for 169 of the 202 uncovered patch lines. The 28 tests that did ship covered only the surrounding adaptors/storages.

Local JaCoCo measurement (`mvn test jacoco:report` for `agent-module/plugins/ktor`, `code.coverage` semantics):

| class | before | after |
|---|---|---|
| `KtorClientSendInterceptor` | 0% (77 lines missed) | **94.8%** (73/77) |
| `KtorClientTraceHolder` | 0% (71) | **100%** (71/71) |
| `KtorPlugin` | 0% (76) | **96.1%** (73/76) |
| `KtorClientParentTraceState` | 95% | 100% |
| `KtorConstants` | 0% | 80% |

Combined: **91.5%** line coverage for the Ktor client package (258/282), **92.3%** including `KtorPlugin`/`KtorConstants` (335/363).

## What was added (4 test classes + 1 test helper, no production changes)

- `KtorClientSendInterceptorTest`: traced calls (header injection verified against a real `HttpRequestBuilder`), unsampled/untraced calls, suspended continuation handling (span finishes but the async context stays open), reentrant sends, recording failures.
- `KtorClientTraceHolderTest`: async finish/cancel CAS semantics, scope ownership (leave-only-when-nested, close-only-when-owner), every failure-swallowing path.
- `KtorPluginTest`: setup transform registrations (enabled/disabled/client-disabled/retransform-disabled) and every transform callback including the two client sender transforms.
- `KtorConstantsTest`: shipped `type-provider.yml` registry consistency.
- `KtorTestServiceTypes`: single test-only `ServiceTypeLocator` populated from the plugin's own `META-INF/pinpoint/type-provider.yml`, so all test classes share one yml-grounded `ServiceTypeProvider` registration instead of duplicating mock registrations. `pinpoint-commons-profiler` is added as a test-scope dependency for `TraceMetadataRegistrar` (same pattern as mssql-jdbc / db2-jdbc).

The lines left uncovered are intentionally defensive: `isDebug` log blocks, a `catch` around code that cannot throw by construction, and the application-type detection branch that needs a real main class.

## Observation worth a follow-up

While covering the reentrant path I noticed `KtorClientSendInterceptor.before()` silently drops a previous `activeState`/`pending` entry without finishing it. With the unit harness I can only verify the documented behavior (`reentrantBeforeDropsPreviousState`); whether the Ktor retry/redirect flow can actually re-enter `execute` on the same thread before `after()` runs is better answered with an agent-level integration test — happy to follow up if it looks worth pinning down.

## Validation

- `./mvnw -pl agent-module/plugins/ktor test` — 75 tests, 0 failures.
- Local JaCoCo numbers above; Codecov/SonarCloud on this PR should show the same classes covered.

### DIFF
--- a/agent-module/plugins/ktor/pom.xml
+++ b/agent-module/plugins/ktor/pom.xml
@@ -27,6 +27,11 @@
             <version>${project.version}</version>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>com.navercorp.pinpoint</groupId>
+            <artifactId>pinpoint-commons-profiler</artifactId>
+            <scope>test</scope>
+        </dependency>
 
         <dependency>
             <groupId>io.ktor</groupId>
--- a/agent-module/plugins/ktor/src/test/java/com/navercorp/pinpoint/plugin/ktor/KtorConstantsTest.java
+++ b/agent-module/plugins/ktor/src/test/java/com/navercorp/pinpoint/plugin/ktor/KtorConstantsTest.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2026 NAVER Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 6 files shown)


jaehong-kim:
Thanks, looks great!

PR #14201, opened 2026-08-26
Scope the realtime active thread panel to the node's own service

jihea-park:
## Summary

- **Send the selected node's service on activeThreadCount.** The servicemap draws nodes belonging to other services, so the realtime active thread panel has to ask the node's own service rather than the screen's one. A WebSocket request carries no header, so the service rides in the request parameters (`serviceName` / `serviceTypeName`, matching `ActiveThreadCountHandler`'s keys). Nothing is sent when `experimental.enableServiceMap` is off, so those installations send the exact same request as before.
- **Keep the pinned active thread target across page moves.** The pinned target lived in a ref while the selected map node lives in a global atom, so leaving the page and coming back left the pin engaged but silently re-pinned to whatever node happened to be selected. The target moved to an atom that remembers the path it was pinned on, so a reconnect resumes the application it was watching and another application's realtime view is unaffected.
- **Select an application, not the service group it sits in.** Left-clicking a collapsed service opened its application list but also marked the group as the current target, sliding open a side panel that draws nothing for a group. Selection now happens only when an application (or a link between two of them) is picked from that popup.
- **Back off before reopening the active thread socket.** The socket reopened the instant it closed, and the backend closes the session when it cannot resolve the `serviceName` this PR starts sending, so connect - request - close could run back to back with nothing in between. Retries now wait 1s, 2s, 5s, 10s, then 30s, resetting on a received message rather than on `open` (a session about to be rejected still opens normally). Leaving the page also fires `close`, which used to reopen a socket from a screen that was already gone.

## Test plan

- [ ] `yarn build` and `yarn test` pass (948 tests).
- [ ] With `experimental.enableServiceMap` **off**: the realtime `activeThreadCount` frame is byte-for-byte what it was before - `{ applicationName }` only.
- [ ] With it **on**: pick a node belonging to another service in the servicemap realtime view; the frame carries that node's `serviceName` / `serviceTypeName`, not the screen's service.
- [ ] Pin on, select a WAS node, move to another page and back: the panel keeps watching the application it was watching before, and the reconnected socket requests that one.
- [ ] Pin off and on again without changing the selection: no duplicate request goes out.
- [ ] Open a different application's realtime view: it starts from that view's own selection instead of the previously pinned one.
- [ ] Servicemap: left-click a collapsed service - the application list opens and the side panel stays as it was; picking `b-1` from the list is what selects it. Same for a link whose end is a collapsed service.
- [ ] Servermap and filteredMap are unchanged (neither response carries `subNodes`/`subLinks`).
- [ ] Kill the backend while the realtime view is open: retries space out instead of hammering, and the panel recovers on its own once it is back.
- [ ] Leave the realtime view: no socket is left reconnecting in the background (DevTools -> Network -> WS).

### DIFF
--- a/web-frontend/src/main/v3/packages/ui/src/atoms/serverMap.ts
+++ b/web-frontend/src/main/v3/packages/ui/src/atoms/serverMap.ts
@@ -88,6 +88,28 @@ export const currentNodeStatisticsAtom = atom<GetHistogramStatistics.Response | 
   undefined,
 );
 
+/**
+ * 실시간 activeThreadCount(WebSocket)의 조회 대상.
+ *
+ * 핀(잠금)이 걸려 있으면 map에서 다른 노드를 골라도 이 대상은 그대로다. 컴포넌트 안에 두면
+ * 다른 화면에 다녀오는 것만으로 사라져(브라우저 탭 전환도 포함된다 — `useTabFocus`가 패널을
+ * 통째로 내린다), 돌아왔을 때 핀은 그대로 꽂혀 있는데 대상만 지금 고른 노드로 바뀐다.
+ * 소켓이 다시 연결되면 떠나기 전에 보고 있던 application을 그대로 이어 보라고 아톰에 둔다.
+ *
+ * `path`는 이 대상을 고른 화면의 경로다. 경로에는 service와 기준 application이 실려 있으므로,
+ * 경로가 다르면 다른 화면을 보고 있는 것이라 고정해 둔 대상도 무효다. (아톰은 전역이라
+ * 화면 remount로 지워지지 않는다. 이 비교가 없으면 다른 application의 실시간 화면에 들어가도
+ * 이전 application의 수치를 보여준다.)
+ */
+export type ActiveThreadTarget = {
+  path: string;
+  applicationName: string;
+  serviceName?: string;
+  serviceType?: string;
+};
+
+export const activeThreadTargetAtom = atom<ActiveThreadTarget | undefined>(undefined);
+
 // server-list 선택시
 export const currentServerAtom = atom<AgentOverview.Instance | undefined>(undefined);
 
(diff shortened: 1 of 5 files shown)


jihea-park:
Review follow-up. Thanks - here is what was applied and what was not.

### Applied

**SonarQube `typescript:S3776` - cognitive complexity 18 > 15 (`AgentActiveThreadFetcher.tsx`).** Fair, and this PR is what pushed it over. Opening the socket, parsing frames, sending the request and the new retry backoff all sat in the component next to the rendering. Moved the socket's whole life into `useActiveThreadSocket`; the component is left deciding what to watch and drawing it. The moved code is unchanged.

### Not applied

**Suppressed comment on line 131 - "can send one stale `activeThreadCount` request before `setStoredTarget` runs".** This does not happen. The effect's dependency is `target` itself, so on the render where the selection changes but `target` has not been updated yet, the dependency is unchanged and the effect does not run at all. It runs on the next render, when `target` already holds the new value, and sends exactly one request for it. The only other way in is a `readyState` change, i.e. a reconnect, where re-requesting the currently stored target is the intent.

**Suppressed comments on lines 262 and 282 - the title and `AgentActiveThreadView` lag one render behind the selection when unpinned.** The lag is real but deliberate. `target` means "the application we are actually requesting", and the title has to name that, not the selection - otherwise the header says B while the numbers below it are still A's. `AgentActiveThreadView` also clears itself whenever `activeThreadCounts.applicationName` differs from the `applicationName` it is given, so feeding it the live selection would blank the chart and hold it blank until the next frame for B arrives, which is a longer inconsistency than the one frame it removes. Using `applicationName` directly is also wrong while the pin is on - not following the selection is the point of the pin.

Worth noting this is already an improvement over the previous code: the target used to live in a ref, and mutating a ref does not re-render, so the title kept the old name until the next incoming frame - roughly a second - rather than one frame.

Build and 948 tests pass on `01c21ae`.

PR #14213, opened 2026-08-31
Find service-grouped applications in the service map search

jihea-park:
## Summary
- A collapsed service is drawn as one group node whose `applicationName` is the service name, so the applications inside it could not be found by name in the map search. `buildServerMapSearchList` now expands a group into the group itself plus its member applications; picking a member centers/selects the group node (the only node on the graph), opens the child list popup and points only the query at that application — the same state as left-clicking the group and picking the child.
- Search rows show the owning service, but only when the node key carries one (3-segment key). Server map keys are 2-segment, so that screen is unchanged. The service name is folded into the cmdk value so equally named applications in different services stay distinct and a service name also matches its members.
- Long names in the group popups and menu titles truncate instead of pushing the popup past the map container, which made the map scroll sideways once the search input took focus.
- Dev mock now emits `outRpcList` on WAS→WAS links only, matching the backend condition the filtered map hint is built from.

## Test plan
- [ ] `yarn build` passes
- [ ] `yarn test` passes (new `buildServerMapSearchList` cases included)
- [ ] Service map (non-DEFAULT service): search a collapsed service by name → group node is centered and its application list opens
- [ ] Search an application that belongs to a collapsed service → group is centered, list opens, and the right panel charts load for that application
- [ ] Search an application drawn as its own node → centered and selected as before, no popup
- [ ] Server map (unchanged path): search list shows `name (serviceType)` with no service column and selection behaves as before
- [ ] Long service/application names: group node and group link popups stay inside the map, and focusing the popup search input does not scroll the map sideways

### DIFF
--- a/web-frontend/src/main/v3/apps/web/dev-mock/mockData.ts
+++ b/web-frontend/src/main/v3/apps/web/dev-mock/mockData.ts
@@ -153,6 +153,16 @@ const makeAppLink = (source: any, target: any, timestamps: number[]) => {
       applicationName: source.applicationName,
       serviceTypeCode: source.serviceTypeCode,
       serviceTypeName: source.serviceType,
+      // outRpcList는 WAS→WAS 링크에만 실린다(백엔드 LinkView#writerFilterHint의
+      // `link.isWasToWasLink()`). 프론트도 양쪽 nodeCategory가 SERVER일 때만 이 값으로
+      // filteredMap hint를 만들므로(useServerMapOnClickMenuItem), 같은 조건으로 내려준다.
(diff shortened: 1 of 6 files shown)


jihea-park:
### Review follow-up

**Copilot review** — reviewed all 6 files and generated no comments, so there was nothing to act on.

**SonarCloud** — reported 1 new issue, applied:

- `typescript:S1077` on `ServerMapCore.tsx:715` — the server type icon in the service group popup had no `alt`. It was flagged as new code only because the line was reformatted in this PR, but the point is fair. The icon repeats what the application name beside it already says, so it now carries `alt=""` and screen readers skip it rather than announcing the image URL.

No other findings. Rebased onto the latest master; build, tests (953) and lint all pass.

PR #14227, opened 2026-09-01
[#noissue] Record suspended Ktor client completion on the resume trace

cj848:
## Summary

Fix the recording of the `KTOR_CLIENT_INTERNAL` completion span for suspended Ktor client calls. When a trace is live at resume time, the span is now recorded directly on that trace instead of binding a new async trace onto the thread-shared coroutine binder. When no trace is live (default configuration), the previous owned-async-trace path is preserved.

## Problem

`KtorClientTraceHolder.finishAsync()` recorded the completion span via `AsyncContext.continueAsyncTraceObject(true)` on the coroutine resume thread. That call binds onto the **thread-shared coroutine binder**, so the result depends on what else happens to be bound there:

- **Foreign-trace nesting.** When the coroutines plugin has already bound its own continuation trace to the binder for the same resume (`profiler.kotlin.coroutines.enable=true`), `continueAsyncTraceObject(true)` returns that foreign trace and the completion span lands inside it as a nested participant. The plugin then has to guess ownership through `ScopeUtils` (a Pinpoint-internal API, not a plugin surface) to decide whether it may close the trace — and that close path could close the coroutines plugin's mid-flight continuation trace, breaking that plugin's own trace bookkeeping.
- Measured with a real agent (coroutines plugin enabled): the completion span of a suspended call whose I/O took ~1.5s appeared inside a foreign continuation trace instead of the caller's trace.

## Fix

- `KtorClientTraceHolder.finishAsync(Throwable)` → `recordCompletion(Trace, Throwable)`:
  - **Live trace at resume time** (coroutines plugin enabled): records the span event **directly on the trace that is live at resume time** (`TraceContext.currentRawTraceObject()`). That trace already carries the right parent chain and is owned by its creator (the coroutines plugin or the original caller), so the holder only pushes a span event onto it and never closes it. The `ScopeUtils` dependency and the foreign-trace close danger are removed.
  - **No trace at resume time** (default configuration: coroutines tracing disabled, the suspended call resumes on the client engine thread): the binder is provably empty, so the holder falls back to `asyncContext.continueAsyncTraceObject(true)` and **exclusively owns** the created async trace — records on it and closes it. Default-configuration behavior is preserved without the `ScopeUtils` ownership guessing. The binder restore (`asyncContext.close()`) runs unconditionally, even if the trace close itself fails.
- `KtorClientContinuationInterceptor` now takes a `TraceContext` constructor argument (standard `addInterceptor` auto-injection; same pattern as the vertx plugin's `HandleExceptionInterceptor(TraceContext)`) and passes `currentRawTraceObject()` to the holder.
- The once-only guard (CAS) and the `asyncContext.finish()` cleanup are unchanged.

### Before / after (suspended call, ~1.5s of I/O, coroutines plugin enabled)

```
Before:                                   After:
ROOT (foreign continuation trace)         ROOT
└─ ...                                      ├─ KTOR_CLIENT execute (7ms)
   └─ KTOR_CLIENT_INTERNAL completion       └─ KTOR_CLIENT_INTERNAL completion (0ms)
      lands inside a foreign trace,            recorded at resume time on the
      and the old close path could              resume trace (never closed here)
      close that foreign trace mid-flight
```

## Verification

- **Unit tests**: 78 pass for the module; holder/interceptor tests were rewritten for the new API (records on the given trace without closing it, fallback owned-trace path with close, once-only CAS, begin/end/close failure paths, binder untouched on the live-trace path).
- **Real-agent measurement** (this change built into a plugin jar, loaded by a real Pinpoint agent, spans sent to a Pinpoint collector; local test app calling a 1.5s slow endpoint):
  - coroutines plugin enabled, launch pattern: `execute` ends at suspension (5ms) and the completion span is recorded on the resume trace 1.55s later, nested under the coroutine `resumeWith` — the live-trace path.
  - default configuration, coroutines plugin disabled, runBlocking pattern: `execute` ends at suspension (5ms) on the caller's trace and the completion span is recorded 1.58s later via the owned async trace fallback — the same visible result as the previous implementation, without the `ScopeUtils` ownership guessing.
  - default configuration, coroutines plugin disabled, fire-and-forget coroutine (no live trace at call time): no span is recorded, unchanged from the previous behavior.
- **Beta deployment**: the live-trace recording mechanism was verified on a beta deployment of a large Kotlin service — 25/25 sampled suspended ktor calls showed the completion span on the caller's trace, with no `UNDEFINED` spans and no corrupted trace trees over a 20+ minute observation window.

---
cc @jaehong-kim — you reviewed the original Ktor client instrumentation PRs (#14090, #14182). This fixes the completion-span recording on top of that work.

### DIFF
--- a/agent-module/plugins/ktor/src/main/java/com/navercorp/pinpoint/plugin/ktor/client/KtorClientContinuationInterceptor.java
+++ b/agent-module/plugins/ktor/src/main/java/com/navercorp/pinpoint/plugin/ktor/client/KtorClientContinuationInterceptor.java
@@ -17,6 +17,8 @@
 
 package com.navercorp.pinpoint.plugin.ktor.client;
 
+import com.navercorp.pinpoint.bootstrap.context.Trace;
+import com.navercorp.pinpoint.bootstrap.context.TraceContext;
 import com.navercorp.pinpoint.bootstrap.interceptor.AroundInterceptor;
 import com.navercorp.pinpoint.bootstrap.logging.PluginLogManager;
 import com.navercorp.pinpoint.bootstrap.logging.PluginLogger;
@@ -24,6 +26,11 @@ import com.navercorp.pinpoint.bootstrap.logging.PluginLogger;
 public class KtorClientContinuationInterceptor implements AroundInterceptor {
     private final PluginLogger logger = PluginLogManager.getLogger(this.getClass());
     private final boolean isDebug = logger.isDebugEnabled();
+    private final TraceContext traceContext;
+
+    public KtorClientContinuationInterceptor(TraceContext traceContext) {
+        this.traceContext = traceContext;
+    }
 
     @Override
     public void before(Object target, Object[] args) {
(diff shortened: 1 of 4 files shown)


jaehong-kim:
Thanks for the PR! The code looks great.

PR #14264, opened 2026-09-08
[#noissue] Inline the map chart query parameters and keep the map root as their base

jihea-park:
## Summary

- The chart hooks compute their query parameters where they are used instead of holding them in state and syncing them in an effect. On the render where the target changed, the parameters were still the previous ones while `serviceName` was already the new one, and that pair formed a queryKey nobody had cached — React Query immediately asked the new service for the previous application.
- `useGetApdexScore` and `useGetHistogramStatistics` take the date range, the map query option and the base application explicitly, so they no longer depend on router state and the screen keeps the decision about what to query.
- The base application of those queries stays the **root of the map**, which is what the statistics API reads it as. The clicked node travels separately in `nodeKey`; the backend compares the two and, when they differ, feeds the root in as that node's from/to peer. Only a node with no usable root falls back to itself — a service map with no application in the path, or a node belonging to another service — and there the identity comes from `nodeKey` instead of the display fields.

## Notes

`HeatmapRealtimeFetcher` keeps its state: `from` samples the `lastToTimestamp` ref that every response rewrites, and reading that ref during render would fire requests outside the five second cadence. Its two states are merged into one so that the parameters and `serviceName` can only change together.

`HeatmapFetcher` keeps reading the URL itself — it is mounted only on non-realtime paths, where no ticking window is involved.

A node's `applicationName` and `serviceType` are display fields, so they cannot be sent as a query identity. A USER node reports the literal `"USER"`, and a database node reports the service type `getDesc()` while the backend resolves `getName()` — `MYSQL` lands on 2100 instead of `MYSQL_EXECUTE_QUERY` 2101, and `MSSQLSERVER` resolves to nothing and answers 400. WAS nodes agree on both, which is why this only shows up on USER and database nodes. `nodeKey` carries the real application name and the `getName()` service type, so the base is derived from it (`parseNodeKeyApplication`).

A USER node is the exception: its own name is synthesised as `{application}_{calleeServiceType}` (`UserNodeUtils#newUserNodeName`) and cannot be split back, because an application name may contain `_` (`IdValidateUtils.ID_PATTERN_VALUE`) and so may a service type name — `my_app_TOMCAT` is ambiguous. It takes the application it calls from the outgoing link's `targetInfo` instead (`findCalleeApplication`).

`/histogram/statistics/links` never reads `applicationName` — it only splits `linkKey`, whose halves are already in the `nodeKey` format — so link selections are left as they were.

## Test plan

- [x] `yarn build`, `yarn test` and `yarn lint` pass
- [ ] Server map: pick an application, click a node and a link — the response time histogram and the server list show that target's numbers
- [ ] Server map: click a `USER` node and a database node — the histogram and the server list are unchanged from before this PR
- [ ] Server map: click a merged node group, then a node inside it — the base falls back correctly and the numbers match the caller's branch
- [ ] Service map (DEFAULT service): click a node that belongs to another service — the charts show that node's numbers instead of coming back empty
- [ ] Service map (non-DEFAULT service): the charts work with no application in the path, including on a `USER` node and a database node
- [ ] Change the period and the map search options (inbound / outbound / wasOnly / bidirectional) — the charts follow, and each change fires one request
- [ ] Switch targets while watching the network panel — no request goes out for the previous application
- [ ] Realtime: the Apdex score and the heatmap keep updating on the five second cadence
- [ ] Filtered map: the Apdex score and the charts are unchanged

### DIFF
--- a/web-frontend/src/main/v3/.claude/rules/service-map.md
+++ b/web-frontend/src/main/v3/.claude/rules/service-map.md
@@ -217,7 +217,7 @@ React Query가 곧바로 요청을 날린다 — 그 service에 없는 applicati
 같은 commit에 배치되므로 짝이 어긋나지 않는다). queryKey와 헤더에는 prop이 아니라 그 state
 (`requestServiceName`)를 쓴다. deferred를 쓰는 스캐터는 serviceName도 같이 deferred로 넘긴다.
 
-→ `useGetScatterData`, `useGetScatterRealtimeData`, `HeatmapFetcher`, `HeatmapRealtimeFetcher`
+→ `useGetScatterData`, `useGetScatterRealtimeData`, `HeatmapRealtimeFetcher`
 
 **파라미터를 그 자리에서 계산하는 훅은 이 처리가 필요 없다** — 애초에 어긋날 렌더가 없으므로
 serviceName도 prop을 그대로 쓴다(`useGetAgentOverview`, `useGetApdexScore`,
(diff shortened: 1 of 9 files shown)


jihea-park:
Verified locally on top of `upstream/master` (no new master commits since the PR base, so nothing to merge): `yarn build`, `yarn test` (1119 tests) and `yarn lint` all pass, and CI is green.

Traced every changed symbol to all of its call sites — each surface is narrow (1-5 usages) and every one was checked:

- `useGetApdexScore`: `disableFetch` was already dead on master (never destructured, no caller passed it), and `dateRange` only moves the `useServerMapSearchParameters()` read from the hook down into `ApdexScoreFetcher`, so the source is unchanged. All five `<ApdexScore>` render sites (server map, filtered map, realtime) keep a valid prop contract.
- `useGetHistogramStatistics`: the new `baseApplication` expression in `ServerMapChartBoard` is equivalent to the `(ignorePathApplication ? undefined : pathApp) ?? fallbackApplication` logic it replaces.
- `getSelectedTargetApplication`: its result only reaches the statistics base, the `nodeKey` fallback and a truthiness guard. Server map is unaffected because the path always carries an application, and with `enableServiceMap` off `useServerMapTargetServiceName` returns undefined, so `isCrossServiceTarget` is always false — existing users never reach the changed path.
- Dropping the state from `HeatmapFetcher` is safe: `parameters` never reaches the chart core, query keys are hashed structurally, and `useGetHeatmapAppData` still guards with `enabled: hasTarget`. `ApdexScore` unmounts its fetcher without a target, and `getQueryString` returns `''` so `enabled: !!queryString` genuinely blocks.

Also checked the backend contract the change rests on. `ServerMapHistogramController#getStatisticsFromServerMap` builds the map from the `applicationName` in the form and compares it against `nodeKey`, feeding the root in as the node's from/to peer when they differ — so keeping the root there is right. `newApplication` splits `nodeKey` on the first `^` with limit 2, matching `parseNodeKeyApplication`, and resolves the type via `findServiceTypeByName`, which is why the display `getDesc()` value cannot be sent. `/histogram/statistics/links` binds the application form but never reads it, so leaving link selections on `sourceInfo` is correct.

Confirmed the fix on a non-DEFAULT service map (`demo-shop-commerce`) by clicking a MySQL node. `MYSQL` is 2100 and `MYSQL_EXECUTE_QUERY` is 2101 with `RECORD_STATISTICS`, and 2101's desc is the string `MYSQL` — so the display field used to resolve to the wrong type and return an empty histogram. The request now carries `serviceTypeName=MYSQL_EXECUTE_QUERY` and the response matches the node's own numbers from the service map payload (Tot 11120, 1s 10682, 3s 110, 5s 121, Slow 207).

One note for a follow-up, not blocking: on a non-DEFAULT service map the node becomes its own root, so the histogram covers every caller, whereas the same node opened from a server map is scoped to the path root. Identical here because the node has a single caller, but they will diverge once a terminal node has several.

PR #14274, opened 2026-09-10
Hide configuration menus without leaving their slot, and take hidden page rules from the consuming app

jihea-park:
## Summary
- Filter `hide` out of the configuration sidebar before rendering: skipping only the render left the gap `space-y-1` had already reserved, so a hidden menu kept an empty slot (and a group whose menus are all hidden now drops the sidebar entirely).
- Turn the single `enableServiceMap` map redirect into one entry of a rule list, and let both the route loader (`resolveHiddenPageRedirect`) and the hook (`useHiddenPageRedirect`) take extra rules, so a page pair that only exists in a consuming build is hidden by the same setting on the way in and in an already open tab.
- Add `createHiddenPagePairRule` for the common fixed-path pair, and drop the "map" part of the names now that the rules are not map-only.

## Test plan
- [ ] `yarn build`
- [ ] `yarn test`
- [ ] Configuration pages: sidebar keeps its spacing, and a group with every menu hidden renders the content full width with no empty slot
- [ ] `/serverMap` with `enableServiceMap` on still redirects to `/serviceMap/DEFAULT`, and the reverse with the setting off
- [ ] Toggle Experimental → Service Map in another tab: the already open map tab still moves to the remaining side


jihea-park:
Both review points were applied in b04e045747.

**Trailing slashes (moderate)** — Applied. The router matches `/config/auth` and `/config/auth/` as the same screen, so the exact comparison in `createHiddenPagePairRule` let a link or bookmark with the slash render the hidden page. The map rule already handled this through `isUnderPage`, so the pair rule was the inconsistent one. It now normalizes the trailing slash on both the pathname and the configured page, and still leaves deeper child paths (`/config/auth/detail`) to whatever owns them. Covered by new tests in both branches — the route loader and the already-open tab — since the two must agree or they bounce a redirect back and forth.

**Parameter names (nit)** — Applied. Renamed `servermapPage`/`servicemapPage` to `pageWhenDisabled`/`pageWhenEnabled`, which is what the helper actually branches on. Map screens have their own rule function, so a pair built by this helper is never a map pair. Argument order is unchanged, so no call site changes.

QA: `yarn build` pass, `yarn test` pass (1153 tests, 4 new), `yarn lint` clean apart from two pre-existing `exhaustive-deps` warnings in `InitialFetchOutlet` that this PR does not touch.

